PR #6, opened 2016-01-14
fix resolving loaders in linked packages

rascada:


template/build/webpack.base.conf.js
```diff
@@ -15,6 +15,11 @@ module.exports = {
       'src': path.resolve(__dirname, '../src')
     }
   },
+
```
yyx990803: Thanks, can you remove the blank lines here? Just keeping things consistent.

rascada: I'm using jscs hence this blank lines.

PR #15, opened 2016-01-29
Try to combin with browserSync

wei-ones:


chrisvfritz:
@wei-ones Sorry it took a while to look at this PR! I haven't used browser-sync before. Can you share more information about it and how it would improve the workflow of everyone using this template?

wei-ones:
@chrisvfritz, thanks for your attention.
It will be a more convenient experience to combine webpack-dev-server with browser-sync, especially when you have to combin with an existing server.
The PR should be seemed as a suggestion, because of lack of adequate testing.

chrisvfritz:
@wei-ones Thanks! We'll investigate this. :smiley:

chrisvfritz:
After some more research, I'm glad I know about browser-sync now, but worry it might not belong in this template. I believe it could be very useful when manually testing across many browsers and devices, but  in apps where it's important to test that different visitors do _not_ see the same thing, it could get in the way. Thank you again though for bringing this tool to our attention @wei-ones!

t2thec:
if anyone else stumbles on this...

I have found that you don't need BrowserSync to test on devices on your network. As the dev server is set up to localhost: `var uri = 'http://localhost:' + port`, you can simply enter your dev machines IP and port to a mobile device and get all the goodness of hot reloading with this webpack setup.

EG: `http://192.168.1.1:8080` Just change the IP to your machine hosting the dev server.

A simple gem that most people probably know, but useful to those who don't.

PR #28, opened 2016-02-24
Add source maps for karma

posva:
Hello,

I just added karma sourcemaps so you can debug easier.

There's no info about how to debug with Phantomjs, I can add it as well of adding how to debug with a real browser (which is much easier) in the readme if you want me to :wink:

### DIFF
--- a/template/build/karma.conf.js
+++ b/template/build/karma.conf.js
@@ -1,5 +1,6 @@
 var webpackConf = require('./webpack.base.conf')
 delete webpackConf.entry
+webpackConf.devtool = 'inline-source-map'
 
 module.exports = function (config) {
   config.set({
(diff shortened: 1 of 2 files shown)


yyx990803:
Thanks! Ran into an issue where it would complain about sourceMap position not found in some cases, tracked down the problem and almost ready to submit a PR to fix it... then found this: https://github.com/karma-runner/karma/issues/1274

Surprisingly it has been there for over a year and was just fixed just 8 days ago...

posva:
That's interesting, I ran into other errors that may be missleading. I'm interested in the code you where running to get those errors. Mind sharing it?

yyx990803:
If you have errors thrown inside a `*.vue` file you will see that warning. (It still works though)

posva:
I see, thanks for the explanation

PR #29, opened 2016-02-26
Improved css loader support

chrisvfritz:
- genericized config for all CSS loaders, so that making them work in both development and production can be done in a single file, while still allowing for differences such as CSS file extraction in production
- explicit support for both `lang="scss"` (unindented syntax) and `lang="sass"` (indented syntax), a la https://github.com/hedefalk/atom-vue/issues/5
- support for both `lang="stylus"` and `lang="styl`"
- all syntaxes have been manually tested in both development and production, so that installing the necessary loader is all that's necessary to get them working

### DIFF
--- a/template/build/css-loaders.js
+++ b/template/build/css-loaders.js
@@ -1,0 +1,52 @@
+var ExtractTextPlugin = require('extract-text-webpack-plugin')
+
+module.exports = function (options) {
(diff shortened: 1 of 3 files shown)


yyx990803:
Awesome! Thanks a lot :)

PR #43, opened 2016-03-13
Add version to input prompts for package.json template

jpokrzyk:
Version number in the package.json file is required per the NPM spec: https://docs.npmjs.com/files/package.json

I only noticed this because I use cmder for cli in windows and it has some sort of built in node parser that complains about it after every command until I add the version number.

I tested and seems to work fine.

### DIFF
--- a/meta.json
+++ b/meta.json
@@ -5,6 +5,12 @@
       "required": true,
       "label": "Project name"
     },
+    "version": {
+      "type": "string",
+      "required": true,
+      "label": "Project version",
+      "default": "0.1.0"
+    },
     "description": {
       "type": "string",
       "required": false,
--- a/template/package.json
+++ b/template/package.json
@@ -1,5 +1,6 @@
 {
   "name": "{{ name }}",
+  "version": "{{ version }}",
   "description": "{{ description }}",
   "author": "{{ author }}",
   {{#private}}


chrisvfritz:
@yyx990803 I hate to say this after-the-fact, but I'm actually wondering whether this shouldn't be reverted. :disappointed: While the docs do say a version is required, [that information is actually out of date](https://github.com/npm/npm/issues/8512#issuecomment-110939142), because it doesn't really make sense for projects that will never be published as a module. So this just creates extra work at the start of a project for no gain.

@jpokrzyk [The real problem is in cmder](https://github.com/cmderdev/cmder/issues/700) and will be fixed in version 1.3. You can follow that link for instructions to get the fix early if you want. :smiley:

yyx990803:
Well I guess it doesn't hurt to have a version, given that you can just press enter for the default value. In the meanwhile this avoids people running into problems with cmder. Another option is removing the prompt but leave the version in there.

jpokrzyk:
I didn't put it in there specifically for me or cmder. That's just how I noticed it. It was a required field per the docs so I thought I should add it. I don't have a strong opinion to have it in there though.

chrisvfritz:
@jpokrzyk You definitely did the right thing. The maintainers of NPM have just chosen to keep a known oversimplification in the docs in an attempt to keep them simpler for package maintainers - at least until they come up with a nice, concise way to explain the exception.

@yyx990803 Given those options, my personal preference would be to skip the prompt and just automatically add a version to `package.json`. I wonder if `"version": "*"` would be a good compromise?

PR #60, opened 2016-03-30
add notes about prerendering content

chrisvfritz:
@yyx990803 Thoughts?


chrisvfritz:
I felt compelled to add a comment explaining the `replace: false`, which could be an indication that it's not worth adding... I'm on the fence.

yyx990803:
I think it's better to keep `App.vue` as-is and add the note in the pre-render section of the README. Since `replace: false` doesn't actually do anything if no pre-render is used.

chrisvfritz:
@yyx990803 Yeah, I think that's a good call. Doc reference looks good?

PR #66, opened 2016-04-03
Move 'static' prefix

pespantelis:


chrisvfritz:
Thanks for the PR @pespantelis! Could you explain what the advantage of this change is? It looks like logic is just moved around a little. Does it work better somehow or prevent other problems?

yyx990803:
Thanks for the change - I do think this change makes sense, although there's one problem: if the user wants to change the sub-directory where built files are placed, they now have to change multiple places in the config instead of one.

Also, the build script relies on the sub-directory name to copy files in `/static` to the right place: https://github.com/vuejs-templates/webpack/blob/master/template/build/build.js#L14

So we probably want to extract the static assets sub-directory name into a variable so that the user only need to change it in one place if they ever need to.

yyx990803:
What we primarily want to consider is how can we make the setup easy to integrate with an existing backend app.

Ideally, this is what the user needs to do in `webpack.prod.conf.js`:

1. Change `output.path` to the root of the backend app;

2. Change the variable mentioned in previous comment (let's call it `staticPath`?) to the sub-directory where static assets are supposed to be placed in for that backend app. E.g. `/public` for laravel/rails apps  (not in `app/assets` because webpack essentially does the job of the asset pipeline).

3. They probably don't have to change `publicPath` unless they want to.

Does this make sense?

pespantelis:
@chrisvfritz @yyx990803 Thanks for your responses :smile: 

@yyx990803 I tried to implement your suggestions.
Please check my solution and let me know your concerns.

yyx990803:
@pespantelis thanks a lot for the follow up - unfortunately, it turns out this problem is more complicated than I thought, and your updates was quite different from what I had in mind. I eventually decided to make some pretty big changes, so sorry for not being able to merge this PR. Really appreciate the effort though.

PR #71, opened 2016-04-08
exclude `build` directory when eslint check

xinchaobeta:
The `build` directory should not be included in the eslint check. Though the current config will not cause any error, but when using eslint-plugin-import will throw error: 

`Unable to resolve path to module 'webpack-hot-middleware/client?noInfo=true&reload=true'`

### DIFF
--- a/template/build/webpack.base.conf.js
+++ b/template/build/webpack.base.conf.js
@@ -36,7 +36,7 @@ module.exports = {
         test: /\.js$/,
         loader: 'eslint',
         include: projectRoot,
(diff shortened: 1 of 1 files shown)


yyx990803:
This is a bug in eslint-plugin-import. You can disable that rule for the file using inline eslint-disable comments.

xinchaobeta:
The file cause eslint throw error is `./build/dev-client.js` 
I think if someone use `npm run dev`, the eslint should not check the `./build` directory

PR #77, opened 2016-04-12
fix static path for windows environment

moonqqqq:
I made this PR with the help of @myst729.

Using back slash (\\) as file separator is how path.join works in Windows environment, which makes loading of static files failed on my windows machine.
img1
In Mac or Linux it should be slash (/), which works.
img2
We think url.resolve is a better approach here.
(The pictures are from @myst729)


yyx990803:
Thanks!

Issue #33, opened 2016-02-29
css-loaders build not works 


zjhiphop:
1. First install 'node-sass' and 'sass-loader'
2. Then in the project main.js, import scss file, like:
```
reqire('path-to/test.scss')
```

Then when i run 'npm run build', there always expose an error: 
```
ERROR in ./src/scss/app.scss
Module parse failed: /Users/jade/Documents/project/workspace/hk/online_jd/wapdev/paotuibang-new/src/scss/app.scss Line 2: Unexpected token ILLEGAL
You may need an appropriate loader to handle this file type.
| /** IMPORTS **/
| @import "./variables.scss";
|
| /** GLOBALS **/
 @ ./src/main.js 31:0-26
```

For simple solution is to add 'scss' loader config inside webpack config file. But how to integrate it with 'css-loaders'?

chrisvfritz:
@zjhiphop Check out [the docs for `sass-loader`](https://github.com/jtangelder/sass-loader#apply-via-webpack-config). You'll need to a loader for SCSS to use them outside of components. The only difference is that instead of the `style` loader, you'll use `vue-style`. Let me know how that works out for you. :smiley:

chrisvfritz:
@zjhiphop Just checking in. Have you had a chance to see if the suggestion above worked for you?

chrispappas:
I am trying to do this literally right now - saw your last comment pop up in real time!

If you can help with an example snippet that would be helpful. I'm trying to do:

```
require('src/assets/scss/main.scss')
```
 inside my App.vue component

chrispappas:
I actually might be on the right track here already, added the following to my `webpack.base.conf.js` file in the `loaders` array (although getting other unrelated errors so I can't quite confirm):

```
      {
        test: /\.scss$/,
        loaders: ["css", "sass"]
      },
```

chrisvfritz:
@chrispappas _Almost_. I believe this would be what you want:

``` scss
{
  test: /\.scss$/,
  loaders: [ 'vue-style', 'css', 'sass' ]
}
```

chrispappas:
Nice, I had actually added the stock style-loader and prepended `style` to the loaders array, which seems to work great. Should I replace it with `vuejs/vue-style-loader` instead? This seems to work just fine for my purposes at this point with the stock `style-loader` plugin.

chrisvfritz:
[...]

PR #81, opened 2016-04-13
add built-in config for standalone compile-to-css files

chrisvfritz:
Allows users to just start coding styles in standalone files with the appropriate extension once they install the appropriate loaders and their peer dependencies, similar to how style elements currently work in `.vue` components.

### DIFF
--- a/template/build/css-loaders-vue.js
+++ b/template/build/css-loaders-vue.js
@@ -1,0 +1,36 @@
+var ExtractTextPlugin = require('extract-text-webpack-plugin')
+
+module.exports = function (options) {
+  options = options || {}
+  // generate loader string to be used with extract text plugin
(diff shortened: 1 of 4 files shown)


chrisvfritz:
@yyx990803 Resolves #33.

yyx990803:
Shouldn't these be extracted in production build too? I think these two should reuse the same dev/prod settings.

chrisvfritz:
Yeah, probably. I only use external files for branding variables that are imported into style elements in `.vue` components. I'll update.

yyx990803:
@Grafikart beat you to it in #84 - sorry Chris ;)

chrisvfritz:
That's fantastic! I'll never complain when someone does work so I don't have to. 😄

PR #84, opened 2016-04-14
Allow external css to be included

Grafikart:
Basically this is the same idea as https://github.com/vuejs-templates/webpack/pull/81 (being able to include CSS outside of .vue files keeping the hot-reload and extract)
But I wanted to rewrite as few lines as possible to do so, and use the work done within `cssLoaders()`

### DIFF
--- a/template/build/utils.js
+++ b/template/build/utils.js
@@ -40,3 +40,17 @@ exports.cssLoaders = function (options) {
     styl: generateLoaders(['css', 'stylus'])
   }
 }
+
+// Generate loaders for standalone style files (outside of .vue)
+exports.styleLoaders = function (options) {
+  var output = []
+  var loaders = exports.cssLoaders(options)
+  for (var extension in loaders) {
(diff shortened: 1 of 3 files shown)


yyx990803:
Nice job!

vaske:
Is there any document which describes where to put scss files and how to include them in app?

chrisbraddock:
http://vuejs-templates.github.io/webpack/pre-processors.html

LinusBorg:
@vaske The comment by @chrisbraddock covers writing SCSS in .vue files. 

If you ask about where toput actual .scss files and how to inlcude those in the app, outside of .vue files, the answers top those two questions are:
* anywhere you want, I would suggest a subfolder in `/src`
* `import './src/scss/your-sass-file.scss'` anywhere in `main.js` will do the job.

vaske:
@chrisbraddock thanks, @LinusBorg yeah this was useful.

PR #100, opened 2016-04-30
Remove dependency on bind

crookedneighbor:
Bind was required for Phantomjs v1, but Phantomjs v2 supports bind natively

If this is desired, I can go ahead and adjust the rest of the templates. Just let me know.


yyx990803:
I think this is the only template that uses Phantom 2, but thanks!

PR #108, opened 2016-05-09
[WIP] Use const and let the syntax of the ES6

vinicius73:
The node-v4 has support for const and let and  the next LTS version (v6) will fully support ES6.
Thinking about it, it is interesting to the node scripts also run in ES6 syntax

There are other possible optimizations, for now I would like feedback from you about this change.

### DIFF
--- a/template/build/build.js
+++ b/template/build/build.js
@@ -1,12 +1,13 @@
+"use strict";
 // https://github.com/shelljs/shelljs
 require('shelljs/global')
 env.NODE_ENV = 'production'
 
-var path = require('path')
-var config = require('../config')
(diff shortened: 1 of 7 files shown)


chrisvfritz:
My personal preference would actually be to avoid this right now. I worry that others might see `let` and `const` and think, "Oh, I can write in ES6!" when in fact, there's only very limited support and will be for the foreseeable future, as we don't plan on dropping support for node-v4 anytime soon.

vinicius73:
Thanks for the feedback.
Anyway, you can close the PR if it considers that this is not the time. :)

Issue #85, opened 2016-04-14
Feature : Autoprefixer


Grafikart:
Since this template is able to load CSS I was wondering if you would be interested into adding autoprefixer.

- PostCSS loader will be used
- browser support would be set using config.js file
- it will be added using `cssLoaders()`

I can do a PR if you are interested.

yyx990803:
vue-loader actually already does autoprefixing for component CSS: http://vuejs.github.io/vue-loader/features/postcss.html

But yeah it doesn't get applied to non-component CSS. But I like your proposal, we can turn off vue-loader's own stuff and centralize it inside `cssLoaders()` and `config.js`.

Grafikart:
Ok I'll work on it tomorrow. I'll close this issue when the PR is submitted

shentao:
@Grafikart how is it going? Just stumbled on this problem myself.

Grafikart:
I did this feature for a project but I didn't add this to the template. Is there a way to test a local template @yyx990803 ? I don't want to make a broken PR :)

chrisvfritz:
@Grafikart I recommend forking the template, then you can run `vue init Grafikart/webpack name-of-test-project` to generate a new project using your fork.

chrisvfritz:
@Grafikart Just pinging on this. Anything we can do to help?

Grafikart:
No it's fine I just did the missing commits

guidobouman:
You can actually import an external stylesheet in your top-most (App) component. That gets parsed as well, right? One implementation to rule them all.

nerdyglasses:
I've added a new entry for other sass files that I don't want to go into a vue component directly. Unfortunately another sass entry doesn't get the autoprefixing. Is that something that can be added easily? I'm trying to figure this out for hours now.

Adding sass directly to the webpack modules hash breaks everything unfortunately.

guidobouman:
That's what they're working on here.

For now you could import it in your topmost component. That way you don't need another entry.

pksunkara:
@Grafikart Do you need any help on this?

guidobouman:
Autoprefixer is in the latest template release. This can be closed.

PR #111, opened 2016-05-11
autoprefixer for external CSS

Grafikart:
Fix feature request #85

Here is a summary : 

- Removed autoprefixer from vue
- Added postCSS loader
- Added postcss autoprefixer
- Added a new configuration setting to define browser support
- Added postcss mqpacker to merge media query

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -5,6 +5,7 @@
 - [Linter Configuration](linter.md)
 - [Pre-Processors](pre-processors.md)
 - [Handling Static Assets](static.md)
+- [Environment Variables](env.md)
 - [Integrate with Backend Framework](backend.md)
 - [API Proxying During Development](proxy.md)
 - [Unit Testing](unit.md)
--- a/docs/env.md
+++ b/docs/env.md
@@ -1,0 +1,51 @@
+# Environment Variables
+
+Sometimes it is practical to have different config values according to the environment that the application is running in.
+
+As an example:
+
+```js
+// config/prod.env.js
+module.exports = {
+  NODE_ENV: '"production"',
+  DEBUG_MODE: false,
+  API_KEY: '"..."' // this is shared between all environments
+}
+
+// config/dev.env.js
+module.exports = merge(prodEnv, {
+  NODE_ENV: '"development"',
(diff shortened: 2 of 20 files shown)


shentao:
@Grafikart Thanks a lot! 🍺

chrisvfritz:
This seems to work. @yyx990803 Does this look like it'll play well with `vue-loader`?

yyx990803:
`vue-loader` still uses postcss internally (when using scoped styles), so I think this would cause postcss to be run twice on Vue component CSS in some cases.

What I think would be more ideal:

1. Only chain `postcss-loader` for non-Vue CSS (probably by adding an option to the `cssLoaders` function)
2. Disable autoprefixer in Vue config, and let both `vue-loader` and `postcss-loader` share the same Array of postcss plugins.

Grafikart:
Ok I see what you mean, I'll try to implement your idea

hekigan:
Hi guys,

I think that this is good stuff, but I am concern with adding ```mqpacker``` to the build process for the following reason:
https://github.com/hail2u/node-css-mqpacker/issues/16

Grafikart:
I got your concern @hekigan, this plugin can be removed if needed after the template is installed so I don't think it's a big deal (but if it's blocking I can remove it)

Grafikart:
@hekigan I removed mqpacker since it was only merging mediaquery from the same file (so it was quite useless to add it :( )

Grafikart:
@yyx990803 I just did the changes you requested

yyx990803:
@Grafikart thanks!
@chrisvfritz would appreciate you taking a look at this

chrisvfritz:
This looks really great. I just submitted [a PR to your PR](https://github.com/Grafikart/webpack/pull/1) 😜 with a slight refactor. I changed the strategy used a little bit, so am happy to hear your thoughts on that. Afterwards, it should be good to merge. 😄

Grafikart:
I'm now wondering if all this work worth it. I mean webpack is not designed for CSS and I think we could achieve the same result doing postcss transform when webpack extract text plugin finish. 

We build => webpack generate assets => postcss is used on compiled CSS (remove duplicates, autoprefix, mergemediaqueries...)

This design could make things simpler.

PR #117, opened 2016-05-14
Support environment variables

pespantelis:
Sometimes we want to have different values based on the environment.

Example:
```js
// prod.env
DEBUG_MODE: false,
API_URL: 'https://example.com'

// dev.env
DEBUG_MODE: true,
API_URL: 'http://localhost:8080'

// usage
Vue.config.debug = process.env.DEBUG_MODE
Vue.http.options.root = process.env.API_URL
```

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -5,6 +5,7 @@
 - [Linter Configuration](linter.md)
 - [Pre-Processors](pre-processors.md)
 - [Handling Static Assets](static.md)
+- [Environment Variables](env.md)
 - [Integrate with Backend Framework](backend.md)
 - [API Proxying During Development](proxy.md)
 - [Unit Testing](unit.md)
--- a/docs/env.md
+++ b/docs/env.md
@@ -1,0 +1,51 @@
+# Environment Variables
+
+Sometimes it is practical to have different config values according to the environment that the application is running in.
+
+As an example:
+
+```js
+// config/prod.env.js
+module.exports = {
+  NODE_ENV: '"production"',
+  DEBUG_MODE: false,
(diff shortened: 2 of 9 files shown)


yyx990803:
Looks good - can you add a section in the docs for this feature?

pespantelis:
Yeah, of course!

pespantelis:
I just added the section to docs. Could you check it? :)

yyx990803:
👍 thanks!

wishinghand:
I'm confused about the usage. Where in a vue-webpack project would I set the Environment variable, 
`Vue.API_URL = process.env.API_URL` for an example like pespantelis originally posted about?

wishinghand:
I figured it out. In the `config/dev.env.js` and `config/prod.env.js` files I added my key:value pair (API_URL: '"URL_GOES_HERE"' and then just referenced them in my .Vue files like this: `process.env.API_URL`. 

For future people who find this thread: pay attention to the double quotes inside of the single quotes. It threw me off for another 20 minutes.

lauhakari:
I have a question here though.
Many of us propably have a dev -stage - production setup. But as the env vars are "hardcoded" in the file, how would I tell the build to use for example a stage.env.js instead of prod.env.js.
Would love it if I could just set some environment variable on the server which the app picks up.

thylux:
I'm trying to use this feature in a project based on this template but process.env is always undefined.
Can anybody check this feature is still working for the latest version of the template?

dtsao:
For future people who find this thread: you have to stop (ctrl-c) and restart the webpack dev server (`npm run dev`) for changes to `config/dev.env.js` to take effect. (I am so used to using HMR, I expected the changes to take effect when I saved my changes.)

PR #118, opened 2016-05-14
Split vendor JS into its own bundle

chrisvfritz:


yyx990803:
Mind doing this for `dist` branch instead? That is what CLI 2.0 is pulling.

chrisvfritz:
Ahh, I see that now. On it.

PR #122, opened 2016-05-16
add vue alias for faster compiling

ryerh:
By default, `import Vue from 'vue'` will redirect to `node_modules/vue/dist/vue.common.js`, which can cause some filesystem search cost. Considering `vue` is frequently imported in projects, so I specific `alias` and `noParse` for `vue` in webpack configuration.

### DIFF
--- a/template/build/webpack.base.conf.js
+++ b/template/build/webpack.base.conf.js
@@ -18,13 +18,17 @@ module.exports = {
     alias: {
       'src': path.resolve(__dirname, '../src'),
       'assets': path.resolve(__dirname, '../src/assets'),
-      'components': path.resolve(__dirname, '../src/components')
+      'components': path.resolve(__dirname, '../src/components'),
+      'vue': path.resolve(__dirname, '../node_modules/vue/dist/vue.common.js')
     }
   },
   resolveLoader: {
     fallback: [path.join(__dirname, '../node_modules')]
   },
   module: {
+    noParse: [
+      /vue\.common\.js/
+    ],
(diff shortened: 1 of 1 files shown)


yyx990803:
This should theoretically save a file read (since webpack no longer needs to read `main` from Vue's `package.json`) and also save the parse time for the entire Vue.js file. The difference is probably not easily noticeable but seems nice to have.

chrisvfritz:
@yyx990803 Should we go ahead and merge this then or is it not worth the added configuration complexity?

yyx990803:
Let's merge as long as it works properly.

chrisvfritz:
@ryerh On a fresh project, I'm getting

```
Uncaught ReferenceError: process is not defined
```

on [this line](https://github.com/vuejs/vue/blob/dev/dist/vue.common.js#L939). Does this PR need an update for the latest `dist` branch?

chrisvfritz:
Huh. Just gave a stab at diagnosing this issue and I can't reproduce it anymore. I'm assuming I somehow did something wrong and this is actually fine, so merging. 😃

chrisvfritz:
Nevermind. Shortly after I merged, someone reported the same issue I did. Looks like it may be an intermittent issue, possibly related to https://github.com/webpack/webpack/issues/2312? Just reverted.

chrisvfritz:
@ryerh Still open to a new pull request if you happen to figure out what causes the issue sometimes. When it was working for me, I did notice faster build times by about ~1.5s on average. 😃

PR #133, opened 2016-05-22
[WIP] Better ESLint config

chrisvfritz:
- Lint the _entire_ codebase instead of just src and tests.
- Rely on ESLint's `--fix` to adjust code style of `.js` files (`.vue` files do not work with `--fix`, so still need to be adjusted manually)
- Add `.eslintrc` files for some subdirectories directories, so that ESLint better understands their context and will not unnecessarily display errors in text editors with ESLint integration.


chrisvfritz:
@yyx990803 This is still a work in progress, but I'd love to hear what you think about this general direction?

yyx990803:
I'm not sure - feels clunky to have so many `.eslintrc` lying around. Although I don't have a better idea at this moment - will take a look later.

chrisvfritz:
My biggest problem with it is that most of those `.eslintrc`s obliterate the project's main linting rules with `"root": true`. I'd love a linting-package-agnostic way of disabling entire classes of rules, with something like this:

``` json
"parserOptions": {
  "ecmaVersion": 5
}
```

But it seems to have no effect. 😞

chrisvfritz:
The most significant pain for AirBnb users has been resolved with the addition of `.eslintignore`, so closing.

PR #134, opened 2016-05-23
Ability to pass arguments to Nightwatch.js process

ianwalter:
Modifying e2e runner script to allow the user to be able to pass arguments to the Nightwatch.js process and override defaults (--config and --env) using the `npm run [cmd] -- [args]` syntax.


yyx990803:
Hey thanks for the PR - can you submit it against the `dist` branch instead?

ianwalter:
Sure, I'm guessing I should have based off of dist as well?

yyx990803:
Yes please

PR #135, opened 2016-05-24
Ability to pass arguments to Nightwatch.js process

ianwalter:
Modifying e2e runner script to allow the user to be able to pass arguments to the Nightwatch.js process and override defaults (--config and --env) using the `npm run [cmd] -- [args]` syntax.


yyx990803:
Thanks!

chlab:
@ianwalter what's the reason for the `--` (`npm run e2e --`) as the first argument? Just trying to understand.

Also, if you guys are interested, I could contribute automatic finding of specific test cases (`npm run e2e --test SomeModule`).

ianwalter:
@chlab The reason for the `--` is so that you pass the arguments to the Nightwatch process instead of NPM when you use `npm run`. Using it, you can pass any argument to Nightwatch, including `--test`.

PR #141, opened 2016-05-27
change to eslint-config-airbnb-base

obingo:
eslint-config-airbnb/base is deprecated, change to eslint-config-airbnb-base


lazystone:
After this change I can't build the project. I get
```bash
ERROR in ./src/main.js

  ✘  https://google.com/#q=import%2Fno-unresolved  Unable to resolve path to module './App'  
  /home/andrey/Downloads/example/src/main.js:2:17
  import App from './App';
                   ^
```

PR #143, opened 2016-05-30
only bundle js files in vendor.js

szrenwei:
There is an issue when `@import` css files from `node_modules`, see [details](https://github.com/vuejs/vue-loader/issues/249).

The vendor.js should only bundle js files.


chrisvfritz:
Thanks! 😄

PR #153, opened 2016-06-04
add beta tag to airbnb linting option

chrisvfritz:
- Added beta tag to Airbnb option so that people aren't surprised when it's not as much a first-class citizen as feross/standard at this point
- Corrected capitalization from `AirBNB` to `Airbnb`

@yyx990803 Does this look OK to you?

### DIFF
--- a/meta.json
+++ b/meta.json
@@ -30,9 +30,9 @@
           "short": "Standard"
         },
         {
-          "name": "AirBNB (https://github.com/airbnb/javascript)",
+          "name": "Airbnb (BETA) (https://github.com/airbnb/javascript)",
           "value": "airbnb",
-          "short": "AirBNB"
+          "short": "Airbnb (BETA)"
         },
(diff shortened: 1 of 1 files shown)


yyx990803:
Hmm not sure if beta makes sense for an option - if Airbnb's new config is still causing trouble, can we just revert to using the old, although deprecated, but maybe still available config?

chrisvfritz:
It's not the new config specifically, but the fact that Airbnb's config isn't really a first-class citizen yet. When generating a new project, many files in the project folder - such as everything in the build folder - don't actually follow the enforced style. This causes a flurry of linting errors in editors with ESLint integration and also just feels yucky. 🙁 We've gotten a few complaints about it, so this was an attempt to better manage expectations.

chrisvfritz:
This has since been resolved by the addition of `.eslintignore`.

PR #169, opened 2016-06-22
Add eslint ignore

funkyLover:
I used to code in st3 with a plugin `eslint`, and I now code with vscode also get a plugin work with eslint. I will get linter error within the editor, I think it is necessary to config the `.eslintignore` file? how do you guys think?


funkyLover:
sorry, I make wrong step with pr....

PR #178, opened 2016-06-27
Fix bug in layout with height bigger than 100%

ivomarsan:
Across (height: 100% + margin: auto)

It's not a big feature. But is my contribution :smile:


chrisvfritz:
Hi @ivomarsan. It looks like something strange is happening with this PR. It's coming from "unknown repository" and seems to include some commits that are already in the repo. 😕

Issue #177, opened 2016-06-27
Dev server content base not working


iltegin:
I tried  this code partials. 
```

var devMiddleware = require('webpack-dev-middleware')(compiler, {
  publicPath: webpackConfig.output.publicPath,
  contentBase:'/',
  stats: {
    colors: true,
    chunks: false
  }
})
```

I am doing wrong?

chrisvfritz:
There's not a lot of information here, but you may want to check out https://github.com/vuejs-templates/webpack/issues/98.

Issue #98, opened 2016-04-28
dev server Cannot GET '/' after changing assetsRoot


plantvsbirds:
I tried to change `build` config and make the project build to different directory(out of project's root directory). Soon after that, I found out that `npm run dev` no longer works and returns `cannot get '/'`.

I managed to solve this by setting `assetsSubDirectory`, `assetsPublicPath`, `assetsRoot` specifically for `dev` environment(and make webpack.dev.conf) to read from it, overwriting the base setting. It's kind of confusing that even in `dev` environment build system reads stuff from `config.build`. 
Would it be better if these `dev` settings come with the teamplate?  I can input a quick PR.

rayzang8:
I met the same issue(file issue #103), and I modify the default configuraion to solve it.

chrisvfritz:
@plantvsbirds I'd be open to taking a look at a PR with what you have in mind.

chrisvfritz:
Haven't heard any reports of this problem for a while, so closing for now - happy to reopen pending new reports or _especially_ a proposed PR that would resolve it. 😄

runarhk:
Experienced same problem. Wanted to use output for an Apache Cordova app running on android. Had to change the assetsPublicPath from '/' to '' to get it running on android device. But when I did this, it would no longer work on dev server. The solution was to fix it like explained in issue #103 . Should the config support two different assetsPublicPaths out of the box?

yupferris:
@chrisvfritz: I've also run into this problem, and as with @runarhk, I was able to fix it as explained by @rayzang8 in #103. The explanation there might not be a PR, but it's pretty close :)

In either case, this still seems to be a problem, and at least one of these tickets should be re-opened until it's resolved.

hekigan:
Actually I have a related issue.

I am completely fine developing in the root directory when in dev environment.
Problem arise when the production folder is in a sub (ex: www.site.com/event/party).

The biggest issue I have right now is the fonts.
The css are parsed with postcss (which is fine), but the `utils.js` that has the `generateLoaders` function kind of prevents me to easily add other plugins to postcss.

My goal was to use something like `postcss-fontpath` to fix the path issue.
In the current situation I can not have a relative path for my fonts. It's always relative to the root folder whatever I do.

Basically it would be great to have the possibility to have all assets relative to the current folder, or being able to specify the absolute path to set in prod environment. One or the other would solve my problem I think.

Anyone else with this kind of problem?

PR #188, opened 2016-07-07
Conditional server path

nickspiel:
This patch will fix an existing issue with dev paths getting mixed up with build paths when defining a different build location.
 
Related issues: #98, #103, #177

### DIFF
--- a/template/build/dev-server.js
+++ b/template/build/dev-server.js
@@ -53,7 +53,7 @@ app.use(devMiddleware)
 app.use(hotMiddleware)
 
 // serve pure static assets
-var staticPath = path.posix.join(config.build.assetsPublicPath, config.build.assetsSubDirectory)
+var staticPath = path.posix.join(config.dev.assetsPublicPath, config.dev.assetsSubDirectory)
 app.use(staticPath, express.static('./static'))
 
 module.exports = app.listen(port, function (err) {
--- a/template/build/webpack.base.conf.js
+++ b/template/build/webpack.base.conf.js
@@ -9,7 +9,7 @@ module.exports = {
   },
   output: {
     path: config.build.assetsRoot,
-    publicPath: config.build.assetsPublicPath,
+    publicPath: process.env.NODE_ENV === 'production' ? config.build.assetsPublicPath : config.dev.assetsPublicPath,
     filename: '[name].js'
   },
   resolve: {
--- a/template/config/index.js
+++ b/template/config/index.js
@@ -19,6 +19,8 @@ module.exports = {
   dev: {
     env: require('./dev.env'),
     port: 8080,
+    assetsSubDirectory: 'static',
+    assetsPublicPath: '/',
     proxyTable: {}
   }
 }


hekigan:
@nickspiel , I tried your fix and it mostly behave as it should.

But when I have fonts in the `static` folder and the build "final" folder is in a sub-directory (ex: `site.com/events/party/static/fonts`), the path in the css is still `site.com/static/fonts`.

hekigan:
Since I did not want to fiddle too much with the core of the template, I wrote an extra file `fixpath.js` in the `build` folder:

```
var fs = require('fs');
var config = require('../config');

function getFiles (dir, files_){
    files_ = files_ || [];
    var files = fs.readdirSync(dir);
    for (var i in files){
        var name = dir + '/' + files[i];
        if (fs.statSync(name).isDirectory()){
            getFiles(name, files_);
        } else {
            files_.push(name);
        }
    }
    return files_;
}

function replaceContent(filePath, stringToReplace, newString) {
	if (!filePath || !stringToReplace || !newString) {
		return console.log('some parameters are missing');
	}

	fs.readFile(filePath, 'utf8', function (err,data) {
	  if (err) {
	    return console.log(err);
	  }
	  var re = new RegExp(stringToReplace,"g");
	  var result = data.replace(re, newString);

	  fs.writeFile(filePath, result, 'utf8', function (err) {
	     if (err) {
	     	return console.log(err);
	     }
	  });
	});
}

// Get all CSS files of the project
var fileList = getFiles(`${config.build.assetsRoot}/${config.build.assetsSubDirectory}/css`);

fileList.forEach((file) => {
	replaceContent(file, `/${config.build.assetsSubDirectory}/fonts`, `${config.build.assetsPublicPath}${config.build.assetsSubDirectory}/fonts`);
});
```

And in `package.json` in the `scripts` part:

```
...
"build": "node build/build.js && npm run fixpath",
 "fixpath": "node build/fixpath.js",
...
```

Not the best solution since I think it would be better to do something about it through `postcss` or something like that, but it does the trick for the moment.

nickspiel:
Sorry I didn't get back to you ssoner, I had a similar problem getting fonts to load (before I had the issue with the build path). I managed to fix it with postcss-import and postcss-url. That portion of mt webpack config looks like the following:

```
vue: {
    loaders: utils.cssLoaders(),
    postcss: [
      require('postcss-import')({
        path: './src/assets',
        addDependencyTo: require('webpack')
      }),
      require('postcss-url')(),
      require('postcss-cssnext')()
    ],
    autoprefixer: false
  }
```
Hope that helps.

hekigan:
It automatically gets the `${config.build.assetsPublicPath}${config.build.assetsSubDirectory}` path from either dev or prod?

nickspiel:
Yeah, that configuration is working for me in both dev and build. Maybe @yyx990803 could provide some better insight? Webpack is still a big black magical box to me.

chrisvfritz:
Thanks for this @nickspiel. Looks good to me! 😃

PR #190, opened 2016-07-09
Updating dependencies

morficus:
Bumping dependencies for the generated project to their latest verions

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -29,36 +29,36 @@
     "connect-history-api-fallback": "^1.1.0",
     "css-loader": "^0.23.0",
     {{#lint}}
-    "eslint": "^2.10.2",
+    "eslint": "^3.0.1",
     "eslint-friendly-formatter": "^2.0.5",
     "eslint-loader": "^1.3.0",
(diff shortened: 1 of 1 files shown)


LinusBorg:
You added a bunch of major version bumps, which means breaking changes. This should be throughly tested.

morficus:
I'm currently using the bumped version on 2 projects w/o any additional config changes.
But I am open to ways of testing these out.

chrisvfritz:
Thanks @morficus! As @LinusBorg mentioned, since these templates are not very modular or easily testable at the moment, a bunch of major version bumps all at once makes me a little uncomfortable, but I would like to bring this in soon - possibly in an "edge" branch for this repo, where it would be more acceptable if something broke temporarily.

To ease development for the future and when some other work for 2.0 slows down a bit, I'd also like to completely rewrite the templates for 2.0 with a full test suite, so that in the future we'll just be able to say, "Tests pass, looks good, merge" for contributions like this.

chrisvfritz:
Closing for now, as we'll either replace this template with a testable one or update dependencies one-by-one.

Issue #123, opened 2016-05-16
CSS source maps not enabled in development environment


PrimozRome:
I have updated to the latest webpack template version and the source maps are not working for me in development environment? Any idea how to solve this issue?

https://dl.dropboxusercontent.com/u/1207859/Screen%20Shot%202016-05-16%20at%2015.23.42.png

Webpack configs

 ```
    var path = require('path')
    var config = require('../config')
    var utils = require('./utils')
    var projectRoot = path.resolve(__dirname, '../')

    module.exports = {
      entry: {
        app: './src/main.js'
      },
      output: {
        path: config.build.assetsRoot,
        publicPath: config.build.assetsPublicPath,
        filename: '[name].js'
      },
      resolve: {
        extensions: ['', '.js', '.vue'],
        fallback: [path.join(__dirname, '../node_modules')],
        alias: {
          'src': path.resolve(__dirname, '../src'),
          'assets': path.resolve(__dirname, '../src/assets'),
          'components': path.resolve(__dirname, '../src/components')
        }
      },
      resolveLoader: {
        fallback: [path.join(__dirname, '../node_modules')]
      },
      module: {
        preLoaders: [
          {
            test: /\.vue$/,
            loader: 'eslint',
            include: projectRoot,
            exclude: /node_modules/
          },
          {
            test: /\.js$/,
            loader: 'eslint',
            include: projectRoot,
            exclude: /node_modules/
          }
        ],
        loaders: [
          {
            test: /\.vue$/,
            loader: 'vue'
          },
          {
            test: /\.js$/,
            loader: 'babel',
            include: projectRoot,
            exclude: /node_modules/
          },
          {
            test: /\.json$/,
            loader: 'json'
          },
          {
            test: /\.scss$/,
            loaders: ['style', 'css', 'sass']
          },
          {
            test: /\.styl$/,
            loader: 'style-loader!css-loader!stylus-loader'
          },
          {
            test: /\.html$/,
            loader: 'vue-html'
          },
          {
            test: /\.(png|jpe?g|gif|svg)(\?.*)?$/,
            loader: 'url',
            query: {
              limit: 10000,
              name: utils.assetsPath('img/[name].[hash:7].[ext]')
            }
          },
          {
            test: /\.(woff2?|eot|ttf|otf)(\?.*)?$/,
            loader: 'url',
            query: {
              limit: 10000,
              name: utils.assetsPath('fonts/[name].[hash:7].[ext]')
            }
          }
        ]
      },
[...]

PR #193, opened 2016-07-11
Add an option to config/index.js to enable CSS Sourcemaps in development

LinusBorg:
### Current behavior

CSS source maps are disabled for development mode in this template because of buggy relative paths in CSS-Loader (https://github.com/vuejs-templates/webpack/issues/123)

But to enable them, one has to manually add set the option in `webpack.dev.conf.js` to true, it involves quite some digging in the webpack config files.

### New Behavior / Improvement

This Pull request adds a `developmentSourceMap` option to `config/index.js`, which is false by default, along with some comments about the css-loader issue described above.

This improves user experience as users can now simply activate CSS sourcemaps from the config file, and are informed about the possibly buggy behaviour instead of stumbling over it after hacking around in `webpack.dev.conf.js`.

```
config: {
  dev: {
    // Sourcemaps off by default because relative paths are "buggy"
    // with this option, according to the CSS-Loader README
    // (https://github.com/webpack/css-loader#sourcemaps)
    // In our experience, they generally work as expexted,
    // just but be aware of this issue when enabling this option.
    // Repated issue in the templates repository:
    // https://github.com/vuejs-templates/webpack/issues/123
    developmentSourceMap: false,
  }
}
```

### DIFF
--- a/template/build/webpack.dev.conf.js
+++ b/template/build/webpack.dev.conf.js
@@ -12,7 +12,7 @@ Object.keys(baseWebpackConfig.entry).forEach(function (name) {
 
 module.exports = merge(baseWebpackConfig, {
   module: {
-    loaders: utils.styleLoaders(sourceMap: config.developmentSourceMap)
+    loaders: utils.styleLoaders({ sourceMap: config.dev.developmentSourceMap })
   },
   // eval-source-map is faster for development
   devtool: '#eval-source-map',
--- a/template/build/webpack.dev.conf.js
+++ b/template/build/webpack.dev.conf.js
@@ -12,7 +12,7 @@ Object.keys(baseWebpackConfig.entry).forEach(function (name) {
 
 module.exports = merge(baseWebpackConfig, {
   module: {
-    loaders: utils.styleLoaders()
+    loaders: utils.styleLoaders(sourceMap: config.developmentSourceMap)
   },
   // eval-source-map is faster for development
   devtool: '#eval-source-map',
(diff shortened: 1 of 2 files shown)


chrisvfritz:
@LinusBorg What would you think about a rename from `developmentSourceMap` to `cssSourceMap`? Since it's under `config.dev`, the "development" part already seems implied to me, but I would like to specify that this only applies to CSS.

I also see a couple more small typos we can fix. 😉 

> // In our experience, they generally work as ~~expexted~~ expected,
> // ~~just~~ but be aware of this issue when enabling this option.

And then I know it was my idea, but what would you think about just removing those last two lines of the comment?

> // Repated issue in the templates repository:
> // https://github.com/vuejs-templates/webpack/issues/123

I feel like linking to css-loader is probably enough when I think about it more. 😃

LinusBorg:
Hi Chris,

1. Already fixed the typos, must found them myself around the time you wrote this. Sometimes I think i'm dyslexic or something :)
2. Renaming it to `cssSourceMap` is of course possible, I simply chose the current name to match the `productionSourceMap` in the prod config settings. Maybe it's more obvious which does what when we leave them like this, with distinctive names?
3. We can of course remove that last line, only added it for you anyway :-P

chrisvfritz:
@LinusBorg Regarding 2, the fact that it's already under the `dev` options scopes it specifically to the development build and [unlike the `productionSourceMap` option](https://github.com/vuejs-templates/webpack/blob/7dc48fa054a2d6e0df30544951bfe38dd37105ce/template/build/webpack.prod.conf.js#L17), this one _only_ affects CSS source maps (not JS), so I think it would be good to make that more clear in the name. 😃

LinusBorg:
Ok, agreed.

Sorry for not having pushed the changes yet, I messed up with a rebase on my end and have to correct it this evening after work.

LinusBorg:
@chrisvfritz I applied all changes as discussed.

PR #197, opened 2016-07-14
Adding Babel register to Nightwatch conf

ianwalter:
This *one weird trick* allows you to use ES6 syntax with your Nightwatch tests!

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -22,6 +22,7 @@
   },
   "devDependencies": {
     "babel-core": "^6.0.0",
+    "babel-register": "^6.0.0",
     "babel-loader": "^6.0.0",
     "babel-plugin-transform-runtime": "^6.0.0",
     "babel-preset-es2015": "^6.0.0",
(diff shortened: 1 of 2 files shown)


chrisvfritz:
`babel-core/register` is actually near deprecation. From the source:

```js
// TODO: eventually deprecate this console.trace("use the `babel-register` package instead of `babel-core/register`");
module.exports = require("babel-register");
```

This file and the surrounding files also follow the [ferross/standard](https://github.com/feross/standard) style, so no unnecessary semicolons please. 😃

ianwalter:
@chrisvfritz Ok, all set.

chrisvfritz:
@ianwalter Thanks! I know this sounds like a nitpick, but in `package.json`, the `babel-register` dependency should be also be placed in alphabetical order. (It's still functional the way you have it, but people are used to scanning their dependencies list with this in mind.)

ianwalter:
Np, nitpicks welcome.

PR #205, opened 2016-07-24
Bugfix/Addition: Actually make the component under test "ready"

Doogiemuc:
The component under test needs to be added to the DOM for it to become ready


chrisvfritz:
@Doogiemuc You're correct that the `ready` hook is not fired, since we're not actually adding this element to the DOM, but the test doesn't require the ready hook to be fired. It works as expected as it is. Did it fail for you?

By the way, the rewrite you proposed is not only slower and more verbose, but I think it would actually fail, since you've now made the test asynchronous without adding a `done` callback to the `it` block.

chrisvfritz:
Closing due to lack of activity.

Issue #210, opened 2016-08-03
HTML linting


ianwalter:
I think adding an HTML linter like [htmlhint-loader](https://github.com/mattlewis92/htmlhint-loader) would be helpful in the same way that eslint is helpful while developing. Thoughts?

chrisvfritz:
@ianwalter This might be useful with an extremely limited set of rules to catch common errors, such as an [unpaired tag](https://github.com/yaniswang/HTMLHint/wiki/Tag-pair). I think I'd like to stay away from any stylistic rules for now though.

ianwalter:
Yea, thats precisely what I was thinking too.

chrisvfritz:
We'd consider a PR then. 🙂

Aladdin-ADD:
now we have eslint-plugin-vue, is this still worthy? @chrisvfritz

chrisvfritz:
@Aladdin-ADD Once eslint-plugin-vue is stable, it'll solve this problem. 🙂

LinusBorg:
I will close this, eslint-plugin-vue is practically around the corner.

PR #224, opened 2016-08-17
Adding #210: HTML linting

ianwalter:


Gomah:
Not a big fan of having ESLint & HTMLHint linked together, wouldn't be preferable to have an extra condition ?

``` json
"htmlLint": {
  "type": "confirm",
  "message": "Use HTMLHint to lint your code?"
},
```

You also need to add an extra filter for the `.htmlhintrc` file:

``` json
"filters": {
  ".eslintrc.js": "lint",
  ".htmlhintrc": "htmlLint",
  "test/unit/**/*": "unit",
  "test/e2e/**/*": "e2e"
},
```

chrisvfritz:
Agreed with @Gomah - I can see cases where someone might want one, but not the other. I very much like that you started with just one rule though, as each one we add probably deserves its own discussion. 😃

PR #225, opened 2016-08-19
Vue 2.0 compat

zigomir:


template/src/App.vue
```diff
@@ -26,6 +26,11 @@
 import Hello from './components/Hello'{{#if_eq lintConfig "airbnb"}};{{/if_eq}}
 
 export default {
+  data () {
+    return {
+      logo: require('./assets/logo.png')
+    }
+  },
   components: {
```
chrisvfritz: The changes in this file should no longer be necessary. They were a workaround for a bug that used to exist in vue-loader, but it's been fixed now.

chrisvfritz:
I'm thinking it would also be good to include a note that people will see when running `vue init webpack my-project`. For example, perhaps another prompt before everything else. @zigomir What do you think?

YamiOdymel:
Should've change `"vue": "^2.0.0-rc.2"` to `"vue": "^2.0.0-rc.3"` now.

template/package.json
```diff
@@ -17,7 +17,7 @@
     "lint": "eslint --ext .js,.vue src test/unit/specs test/e2e/specs"{{/lint}}
   },
   "dependencies": {
-    "vue": "^1.0.21",
+    "vue": "^2.0.0-rc.2",
```
YamiOdymel: it's now `"vue": "^2.0.0-rc.3"`

template/src/App.vue
zigomir: Thanks! Fixing this now.

zigomir:
@chrisvfritz yeah, it certainly can be done. I'll prepare something soon and ping you for feedback.

chrisvfritz:
@zigomir Yeah, it's a tough balance. 😕  Part of me would like to only support the Standard config, but the AirBnB config _is_ much more popular. Supporting it does make maintenance a bit more difficult, but it also means we don't have to respond to many issues complaining about a lack of support for it - I'm not sure which is more of a pain. 😃

chrisvfritz:
@zigomir Just realized, since the new vue-loader no longer includes autoprefixing, we'll also need to add that to the base build with:

``` js
vue: {
  postcss: [
    require('autoprefixer')({
      browsers: ['last 2 versions']
    })
  ]
}
```

Along with this, `autoprefixer` will have to be added to the `devDependencies`.

zigomir:
@chrisvfritz thanks! I went and edited `webpack.base.conf.js` so `vue` key looks like so:

```js
vue: {
    loaders: utils.cssLoaders(),
    postcss: [
      require('autoprefixer')({
        browsers: ['last 2 versions']
      })
    ]
  }
```

What's funny/strange though, is that `webpack` won't complain at all even though I didn't add `autoprefixer` to `package.json` at all. I don't get that, do you?

Also, this branch breaks `hot-reload` and I wasn't able to figure out why ... because I'm a `webpack` newbie + I kinda started hating on it while trying to make hot reload work. So this is my call for help if someone else wants to pick this `2.0` template up because I'm really not motivated fighting webpack :)

chrisvfritz:
@zigomir It's because `css-loader` depends on `cssnano` which depends on `autoprefixer`. (To figure that out, I ran `npm list | grep -B 10 autoprefixer`, which looks for autoprefixer in our dependency tree and shows that line and the 10 before it.) Since this is really an implementation detail though (`cssnano` might update to no longer depend on `autoprefixer` or `css-loader` might update to no longer use `cssnano`), it's good for us to explicitly include it in our devDependencies too.

I'm also happy to investigate the hot reload stuff. 😃  I've already spent a lot of time hitting my head against the Webpack wall, so may as well get some use out of it!

chrisvfritz:
Hmm, hot reloading seems to be working for me. So... I guess I fixed it? 😜

zigomir:
Huh, it works for me now too :) Have no idea what happened in between. I added `autoprefixer` now. Thanks for detailed explanation! :)

chrisvfritz:
No problem - and feel free to let me know if you come across more Webpack-specific things I can help with. I'm happy to save you the frustration if I can. 🙂

niallobrien:
Hi all, is this good to merge yet?

LinusBorg:
I just added a couple ofg things to the `2.0` branch, and it works fine.

But we still should add a prompt/warning that this template now uses 2.0 by default - that is still missing.

zigomir:
We just need to remove this if and we'll have it ;)

https://github.com/vuejs/vue-cli/blob/master/bin/vue-init#L109

niallobrien:
I'm using RC5 with the webpack config with no issues so far. :)

meta.json
```diff
@@ -15,6 +15,22 @@
...
       "message": "Author"
     },
+    "build": {
+      "type": "list",
+      "message": "Vue comes in two build versions, which do you want to use?",
+      "choices": [
+        {
+          "name": "Runtime-only: lighter, but no support for templates defined in .html files (.vue files are fine)",
+          "value": "runtime",
+          "short": "tuntime"
```
jvanbrug: `tuntime` typo?

LinusBorg: ...indeed.

zigomir:
@chrisvfritz I just tried this branch and to me it all looks and works fine. We only need to decide into which branch we want to merge it (#270). I'm for `master` ;)

marines:
Webpack generates sourcemaps in production build. Does it have any purpose?

chrisvfritz:
I agree with you. 🙂 master is my vote.

chrisvfritz:
@marines That comment seems unrelated to this thread 😕 , but yes it does. It allows you to get  stack traces with files and line numbers when debugging an error that happens in production.

zigomir:
Closing in favor of #281.

PR #255, opened 2016-09-11
[2.0] add nightwatch global with devServer's defined URL:Port (fix #241)

### LINKED ISSUE #241: Hardcoded url port breaks e2 tests when devServer port is changed in /config.
The server port is [configurable](https://github.com/vuejs-templates/webpack/blob/1.0/template/build/dev-server.js#L11) by environment variables and config files.
However the e2e tests have a [hardcoded port](https://github.com/vuejs-templates/webpack/blob/1.0/template/test/e2e/specs/test.js#L7), so they break when the port is changed. 

Any ideas on how to use the same port without repeating the logic from the configuration?

LinusBorg:
Fix for #241 

This adds the devServer URL as a global and uses the port defined in `/config/index.js` for the URL, _EDIT:_ or, if provided, a PORT environment variable.

in the demo spec file, this global is then used as the target url for the test instead of hardcoding `http//localhost:8080`

### DIFF
--- a/template/test/e2e/nightwatch.conf.js
+++ b/template/test/e2e/nightwatch.conf.js
@@ -23,7 +23,7 @@ module.exports = {
       "selenium_host": "localhost",
       "silent": true,
       "globals": {
-        "devServerURL": "http://localhost:" + config.dev.port
+        "devServerURL": "http://localhost:" + (process.env.PORT || config.dev.port)
       }
     },
 
--- a/template/test/e2e/nightwatch.conf.js
+++ b/template/test/e2e/nightwatch.conf.js
@@ -23,7 +23,7 @@ module.exports = {
       "selenium_host": "localhost",
       "silent": true,
       "globals": {
-        "devServerURL": "http://localhost" + config.dev.port
+        "devServerURL": "http://localhost:" + config.dev.port
       }
     },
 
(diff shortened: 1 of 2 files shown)


aristidesfl:
Is it possible in this version to specify a port via env variable?

LinusBorg:
As described, it uses the port from the config file, not from ENV

aristidesfl:
Wouldn't being able to specify the port from env be desirable? The use case is being able to easily run tests in a different port than the one used for dev, without having to mess a lot with the configuration. This would allow you to run the dev server at the same time as the tests for example.

LinusBorg:
This is a nice idea, but was not what I understood as the requested functionality from your issue #241 .

Can you open a new issue for this? As the e2e tests use the dev server as well, we will have to change server.js to support both at the same time with different ports.

aristidesfl:
The code I referenced #241 supports both dev config and env variables: https://github.com/vuejs-templates/webpack/blob/1.0/template/build/dev-server.js#L11

Has this been removed in more recent versions? I didn't fully understand what is the relationship of the branch `dist` with `1.0` and `2.0`.

LinusBorg:
No, this still exists in the 2.0 branch (which is identical to the 1.0 branch except for changes nessessary to make it work with 2.0, which are very few).

The point is, it has to be added to the changes in nightwatch.conf.js of this PR as well. Can do that right now I guess.

LinusBorg:
Done

aristidesfl:
Great, thanks! 👍

chrisvfritz:
LGTM!

Issue #242, opened 2016-09-06
[2.0] Add init option to select between standalone and runtime-only


LinusBorg:
I see many people in gitter and the forums stumbling over the fact that importing the standalone version like `import vue from 'vue/dist/vue'` will screw up vue-loader / vueify.

We already explained why and how to prevent that here in the Vue Wiki:

https://github.com/vuejs/vue/wiki/Vue-2.0-RC-Starter-Resources#standalone-vs-runtime-builds

But of course this information is not reaching everyone out there.

I think we should therefore add a warning comment ion main.js with either a link to the wiki section above or short instuctions how to add an alias in the conf for webpack/vueify.

Will work on this.

Sibling ticket for the browserify template [#12](https://github.com/vuejs-templates/browserify/issues/12)

LinusBorg:
Idea: Maybe we should even add it as an option to the install process: 

`"Use standalone instead of runtime-only build?" yN`

chrisvfritz:
Having all the templates pre-compiled will also offer better performance.

LinusBorg:
added PR #256

PR #256, opened 2016-09-11
Add select option to init process: Standalone or runtime-only? (solve #242)

LinusBorg:
Solves #242

User can select between the two build versions:
* runtime-only
* standalone

bildschirmfoto 2016-09-11 um 21 25 16

* If "runtime-only" is seleted, nothing happens as this is the default import
* If "standalone" is selected, we add an alias to the webpack conf and a comment in main.js informing the user about this.

### DIFF
--- a/template/src/main.js
+++ b/template/src/main.js
@@ -1,4 +1,4 @@
-{{if_eq build "standalone"}}
+{{#if_eq build "standalone"}}
 // the following line loads the standalone build instead of Runtime-only
 // thanks to an alias defined in /config/webpack.base.conf.js
 // So you don't have to do import Vue from 'vue/dist/vue'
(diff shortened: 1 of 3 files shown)


template/src/main.js
```diff
@@ -1,3 +1,9 @@
+{{#if_eq build "standalone"}}
+// The following line loads the standalone build of Vue instead of the runtime-only build,
+// so you don't have to do: import Vue from 'vue/dist/vue'
+// (also, importing Vue standalone this way breaks vue-loader, so don't do it)
+// This is done with an alias defined in /config/webpack.base.conf.js
```
chrisvfritz: Should this be /build/webpack.base.conf.js?

LinusBorg: Oh, of course it should.

LinusBorg: Corrected.

PR #257, opened 2016-09-11
[1.0] add nightwatch global with devServer's defined URL:Port (fix #241)

### LINKED ISSUE #241: Hardcoded url port breaks e2 tests when devServer port is changed in /config.

LinusBorg:
Same as #255, but cherry-picked for 1.0 branch

### DIFF
--- a/template/test/e2e/nightwatch.conf.js
+++ b/template/test/e2e/nightwatch.conf.js
@@ -23,7 +23,7 @@ module.exports = {
       "selenium_host": "localhost",
       "silent": true,
       "globals": {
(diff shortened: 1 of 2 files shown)


chrisvfritz:
LGTM!

PR #265, opened 2016-09-18
multi-html and svg support

bluedarker:


LinusBorg:
Hi @bluedarker,

thanks for this PR. This looks very interesting but at the same time, makes the template config even more complex.

We will have to discuss weither this is something we want.

PS: it would help if you added the two things (mutli-HTML vs. SVG-sprite) in two seperate PRs)

PR #268, opened 2016-09-19
Optimize babelrc

dnshi:
This is an update to babelrc. 

Most developers when building a web app are using a modern browser and most of modern browsers have much better supports to ES2015 (even better than babel: https://kangax.github.io/compat-table/es6/). Therefore there is no need to translate code to ES2015 in development mode. 

We still need `transform-es2015-modules-commonjs` since webpack and browsers does not support ES6 commonjs modules.

### DIFF
--- a/template/.babelrc
+++ b/template/.babelrc
@@ -1,5 +1,13 @@
 {
-  "presets": ["es2015", "stage-2"],
-  "plugins": ["transform-runtime"],
+  "env": {
+    "development": {
+      "presets": ["stage-2"],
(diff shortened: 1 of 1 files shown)


chrisvfritz:
I don't think this is a good idea, as the transpiled code sometimes behaves slightly differently from native implementations - especially in terms of performance characteristics. There would also be no way to test real behavior for older browsers in development.

dnshi:
This babelrc config depends on developers. Some developers don't care about the differences and older browsers and some others would take it as a consideration. 

Thanks. I will close this PR.

PR #273, opened 2016-09-23
Change webpack alias to work with standalone default export.

LinusBorg:
This should be merged if/when the default export of the Vue 2.0 package is changed to the standalone build.

### DIFF
--- a/template/build/webpack.base.conf.js
+++ b/template/build/webpack.base.conf.js
@@ -16,9 +16,11 @@ module.exports = {
     extensions: ['', '.js', '.vue'],
     fallback: [path.join(__dirname, '../node_modules')],
     alias: {
-      {{#if_eq build "standalone"}}
-      // nessessary to make "import Vue from 'vue'" load the standalone version.
+      {{#if_eq build "standalone"}}∞
       'vue': 'vue/dist/vue',
+      {{/if_eq}}
+      {{#if_eq build "runtime"}}∞
+      'vue': 'vue/dist/vue.common.js',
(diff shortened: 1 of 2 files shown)


template/build/webpack.base.conf.js
```diff
@@ -16,10 +16,12 @@ module.exports = {
     extensions: ['', '.js', '.vue'],
     fallback: [path.join(__dirname, '../node_modules')],
     alias: {
-      {{#if_eq build "standalone"}}
-      // nessessary to make "import Vue from 'vue'" load the standalone version.
+      {{#if_eq build "standalone"}}∞
       'vue': 'vue/dist/vue',
       {{/if_eq}}
```
chrisvfritz: Won't the above 3 lines be unnecessary if the standalone version is the default export?

LinusBorg: They will, technically. I thought I leave them in to make things clear and not take any chances of it exporting the runtime-only for some people depending on when all the 2.0 goes live in non and reaches the users. 

But it's not nessessary.

chrisvfritz: That's a good point! Then maybe we should merge this even before a decision is made, since it future-proofs users of the 2.0 branch against any possible changes of the default export.

LinusBorg: Agreed

chrisvfritz:
@zigomir What do you think? Good to merge?

zigomir:
Yeah, for sure. I also always favor explicitness 👍

PR #287, opened 2016-09-30
Some extra cleaning

pksunkara:


pksunkara:
@chrisvfritz Do I still need [this](https://github.com/vuejs-templates/webpack/blob/master/template/config/test.env.js) file when I haven't asked for tests to be added. I wasn't sure if I need to remove it.

chrisvfritz:
If it doesn't break the other NPM scripts, I'd be OK making it conditional.

pksunkara:
@chrisvfritz I am not sure what will break because of them which is why I was asking you.

pksunkara:
I am more concerned about why we are checking for testing environment in production configuration file. https://github.com/vuejs-templates/webpack/blob/master/template/build/webpack.prod.conf.js#L10

chrisvfritz:
I believe at one point E2E tests used a superset of the production config, so I don't think that line should be necessary anymore since they're based off the base config now. Unfortunately, the way to test whether the NPM scripts will still work is to just test them manually right now.

It's on my list to refactor (or possibly rewrite) this template and add build tests, so that we can say with a greater degree of certainty whether something will have unintended side effects or not. Until then we do our best. 😅

pksunkara:
I think E2E tests are still using it. Check these 3 lines, https://github.com/vuejs-templates/webpack/blob/master/template/test/e2e/runner.js#L1-L3

chrisvfritz:
Ah, yep - didn't find it in a quick search earlier.

pksunkara:
But the issue is karma tests will also be somehow using this and I didn't find a way to do `{#unit or #e2e}` which is why I am looking for some help on this.

chrisvfritz:
Oh, OK - I see what you're asking now. That can be accomplished with a [custom handlebars helper](https://github.com/vuejs/vue-cli/blob/master/README.md#custom-handlebars-helpers) - specifically, something [like this](http://stackoverflow.com/questions/8853396/logical-operator-in-a-handlebars-js-if-conditional).

pksunkara:
@chrisvfritz Here's the PR then. https://github.com/vuejs-templates/webpack/pull/294

PR #296, opened 2016-10-04
Added vue-router

pksunkara:


pksunkara:
@chrisvfritz Would you be reviewing this?

chrisvfritz:
I took a quick look on my phone and it looks good from a quick glance! I'll try to do a more thorough review later today or tomorrow.

pksunkara:
I am currently adding Vuex to this too.

chrisvfritz:
To keep things simple, I'd actually prefer a separate PR after this for Vuex.

pksunkara:
Okay then. This is ready to be merged. I would add Docs in a separate PR before the next release.

pksunkara:
By the way, please use "Create a Merge Commit" option when merging the PR. The "Squash and Merge" option is not contributor friendly and makes serial contributing more complex than needed.

template/src/App.vue
```diff
@@ -1,35 +1,48 @@
 <template>
   <div id="app">
     <img class="logo" src="./assets/logo.png">
+    {{#if router}}
+    <router-view></router-view>
```
chrisvfritz: Hmm... this is a tough situation. On the one hand, if we're going to include Vue Router, I'd like to also set up some scalable conventions and demonstrate more router features, like including navigation with `<router-link>` above. As it is, `<hello>` is also a bit silly as a page-level component. I'd prefer a `<home>` component and one other component - maybe `<about>`. I also typically recommend that people only use the `components` folder for _reusable_ components, rather than page-level components. I think it's better to have `pages` and `layouts` directories, where wrapper components from `layouts` are used as the root for page-level components in `pages`. 

Adding all this would mean a great deal more complexity for this template however. @zigomir @LinusBorg What are your thoughts? I'm worried that with this proposal, the people who would really benefit from out-of-the-box Vue Router integration would still be left with most integration questions unanswered, as they could probably get this far just from copying and pasting from the Getting Starting guide in the router docs. But adding a more complete Vue Router integration would be a maintenance nightmare.

Maybe we should just have a separate template that includes Vue + Vue Router + Vuex?

pksunkara: I had the same thought but realised we should start the discussion somehow.

template/src/router.js
```diff
@@ -0,0 +1,17 @@
+import Vue from 'vue'{{#if_eq lintConfig "airbnb"}};{{/if_eq}}
+import VueRouter from 'vue-router'{{#if_eq lintConfig "airbnb"}};{{/if_eq}}
+import { Hello } from './components'{{#if_eq lintConfig "airbnb"}};{{/if_eq}}
```
pksunkara: @chrisvfritz This import is not working. Can you check it once?

I had to do the following to get it working:

```js
import Hello from './components/Hello';
```

template/src/App.vue
pksunkara: To be honest, this is a good enough start with the vie-router integration. Sure, Hello is a placeholder component, but the `App` component is setup and people can replace the Hello component.

But I also agree with the complete setup path. Adding a template with vue-router, vuex, vue-mdl.

zigomir: I agree with @chrisvfritz. I think we should keep official vue templates as simple as possible. If we are going to add `router` and `vuex`, I'd definitely do it as a separate template. Although on the other hand this means lot's of copy/pasta boiler plate thats already here (all webpack config and stuff).

I'd use https://github.com/egoist/vuepack if I'd need `full-stack-vue` today :)

I don't know, I'd keep it as simple as possible. If adding `vue-router` or `vuex` is hard to this template, I'd start fixing router or vuex (or their docs) instead.

pksunkara:
Any update on this?

LinusBorg:
Please don't take this too seriously, but I  just came across this tweet a day ago:

https://twitter.com/slicknet/status/782274190451671040

pksunkara:
Well, it depends on what we are dealing it. Here, a decision needs to be made by the Vue core team on whether to follow this route or not. And sometimes, these are not done on github but on other channels of communication.

pksunkara:
Looks like I will be closing this. If anyone wants an extended web pack template with router, vuex etc.., check out https://github.com/pksunkara/spoiler

gkatsanos:
@pksunkara would you mind having a look at my repo/branch? I am trying to integrate your proposed changes to this template (so, having vue-router) : https://github.com/gkatsanos/gplus/tree/redesign

pksunkara:
@gkatsanos You can directly use the repo I mentioned above. I have all these changes and more in my template.

pksunkara:
And also, I am regularly keeping it up-to-date with this template.

gkatsanos:
@pksunkara thank you. I am mostly trying to make it work on my own and understand a bit how it functions.. (I was expecting the integration to be smoother)

PR #300, opened 2016-10-06
Resolve imports

guidobouman:
This PR adds eslint-import-resolver-webpack so AirBnB's config can resolve imports.

Just wondering, do we want import to not require the .vue extension as well?

I could commit this as well:
```
    {{#if_eq lintConfig "airbnb"}}
    // don't require .vue extension when importing
    'import/extensions': ['error', 'always', {
      'js': 'never',
      'vue': 'never'
    }],
    {{/if_eq}}
```

### DIFF
--- a/template/.eslintrc.js
+++ b/template/.eslintrc.js
@@ -15,6 +15,16 @@ module.exports = {
   plugins: [
     'html'
   ],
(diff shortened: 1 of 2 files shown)


chrisvfritz:
It sounds like you're adding this to your other PR, so we can leave it out here.

PR #311, opened 2016-10-10
Fix vue resolve aliases

LinusBorg:
1. I removed the alias for the runtime-build as it's not nessessary anymore (It was only a temporary security measure for the time of 2.0 release)
2. improved alias for the standalone build:
```js
//before
vue: 'vue/dist/vue' // =Y this made it impossible to import other files from the /dist directory.

// now
'vue$': 'vue/dist/vue' // Whith this match pattern, other requires are still possible.
```


LinusBorg:
@chrisvfritz We should update the hint on the "installation" page of the guide accordingly.

pksunkara:
The commits are messed up here.

LinusBorg:
Damn, sorry. 

My fork must have gotten messed up when I tried to merge to upstream after `2.0` got merged into `master`

Will have to re-do this tomorrow.

PR #312, opened 2016-10-10
Improve loader config

LinusBorg:
1. Added more comments so people might be able to find their way better
2. added NODE_ENV 'development'  to devServer.js
3. when `cssSourceMaps` is defined in`config.js`, pass the option not only to first-level CSS-loaders, but also those for vue-loader. Without this, source maps are not generated for CSS from components.


LinusBorg:
Seems I was missing some recenrt commits from the github repo?

PR #314, opened 2016-10-11
fix vue alias

LinusBorg:
1. I removed the alias for the runtime-build as it's not nessessary anymore (It was only a temporary security measure for the time of 2.0 release)
2. improved alias for the standalone build:
```js
//before
vue: 'vue/dist/vue' // =Y this made it impossible to import other files from the /dist directory.

// now
'vue$': 'vue/dist/vue' // Whith this match pattern, other requires are still possible.
```

### DIFF
--- a/template/build/webpack.base.conf.js
+++ b/template/build/webpack.base.conf.js
@@ -17,10 +17,7 @@ module.exports = {
     fallback: [path.join(__dirname, '../node_modules')],
     alias: {
       {{#if_eq build "standalone"}}
-      'vue': 'vue/dist/vue',
-      {{/if_eq}}
-      {{#if_eq build "runtime"}}
-      'vue': 'vue/dist/vue.common.js',
+      'vue$': 'vue/dist/vue',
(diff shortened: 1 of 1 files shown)


pksunkara:
Shouldn't you be removing the `runtime` option from `meta.js` too?

LinusBorg:
No. We only want to add that alias to the webpack config if the user wants to use the runtime-only build. 

The question could be changed into a "yes/no" question, but I think it is working just fine this way.

pksunkara:
So, If I want to require runtime library, I still do `import Vue from 'vue';`? Or should I do `import Vue from 'vue.common';`?

LinusBorg:
The runtime build is the default export of the vue package, so you simply do `import Vue from 'vue'`.

If you select the standalone build, we add the alias, so `import Vue from 'vue'` nmow gives you standalone instead of runtime-only.

Does that help?

pksunkara:
Ah, understood now.

PR #315, opened 2016-10-11
Source maps option for vue css loaders

LinusBorg:
#### The Problem
Some time ago, I added a `cssSourceMap` option to the dev config in `config/index.js`.

Hwever, I used this config only [here, in webpack.dev.con.js](https://github.com/vuejs-templates/webpack/blob/master/template/build/webpack.dev.conf.js#L15), and not [here, in webpack.base.conf.js](https://github.com/vuejs-templates/webpack/blob/master/template/build/webpack.base.conf.js#L86)

Result: CSS sourcemaps were not created for vue files, only for external files(.scss) required in JS.

### Solution

add this option also in that second place. easy.

Well the challenge was that the css loaders for vue.loader are added in `webpack.base.conf.js`, so I had different environments into account. 

That's why I also explicitly set `process.env.NODE_ENV` in `devServer.js` (unless already defined) - so that I can cleanly check the env in the base config.

If someone has a better solution for this, go ahead! :)

### DIFF
--- a/template/build/dev-server.js
+++ b/template/build/dev-server.js
@@ -1,3 +1,5 @@
+if (!process.env.NODE_ENV) process.env.NODE_ENV = 'development'
+
 var path = require('path')
 var express = require('express')
 var webpack = require('webpack')
@@ -7,8 +9,6 @@ var proxyMiddleware = require('http-proxy-middleware')
 var webpackConfig = {{#if_or unit e2e}}process.env.NODE_ENV === 'testing'
   ? require('./webpack.prod.conf')
   : {{/if_or}}require('./webpack.dev.conf')
-
-if (!process.env.NODE_ENV) process.env.NODE_ENV = 'development'
 
 // default port where dev server listens for incoming traffic
 var port = process.env.PORT || config.dev.port
--- a/template/build/dev-server.js
+++ b/template/build/dev-server.js
@@ -8,6 +8,8 @@ var webpackConfig = {{#if_or unit e2e}}process.env.NODE_ENV === 'testing'
   ? require('./webpack.prod.conf')
   : {{/if_or}}require('./webpack.dev.conf')
 
+if (!process.env.NODE_ENV) process.env.NODE_ENV = 'development'
+
 // default port where dev server listens for incoming traffic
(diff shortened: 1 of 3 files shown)


template/build/utils.js
```diff
@@ -25,6 +25,8 @@ exports.cssLoaders = function (options) {
       return loader + (options.sourceMap ? extraParamChar + 'sourceMap' : '')
     }).join('!')
 
+    // Extract CSS when that option is specified
+    // (which is the case during production build)
```
LinusBorg: Note: this is not *really* related to this PR, just some added commenting, might help someone reading it.

chrisvfritz:
@LinusBorg Is this still work-in-progress or ready for review?

LinusBorg:
It's ready

template/build/dev-server.js
```diff
@@ -1,3 +1,5 @@
+if (!process.env.NODE_ENV) process.env.NODE_ENV = 'development'
```
chrisvfritz: I'm thinking since it's the dev server, `NODE_ENV` should always be overwritten here. Could we also place it a little lower and use [the value from the config](https://github.com/vuejs-templates/webpack/blob/master/template/config/index.js#L20)? For example:

``` js
process.env.NODE_ENV = config.dev.env
```

template/build/utils.js
chrisvfritz: 👍

template/build/dev-server.js
LinusBorg: The sec server is also used for testing. Is it wise to override the 'testing' environment?

chrisvfritz: Good point. I still think it'd be a good idea to set to `config.dev.env` though, instead of directly to `'development'`.

LinusBorg: agreed

chrisvfritz: @LinusBorg Just pinging on this. I think once we update this one bit, it should be good to merge. 🙂

LinusBorg: @chrisvfritz  Done

PR #321, opened 2016-10-17
Update default build and descriptions

chrisvfritz:
Since we've seen a lot of people struggle with the runtime-only build, I thought it might be good to switch the default to the standalone version. I've also updated the build descriptions to (hopefully) be clearer and I'd love feedback on whether I've been successful. 🙂

### DIFF
--- a/meta.js
+++ b/meta.js
@@ -26,17 +26,17 @@ module.exports = {
     },
     "build": {
       "type": "list",
-      "message": "Vue comes in two build versions, which do you want to use?",
+      "message": "Vue build",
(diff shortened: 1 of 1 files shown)


chrisvfritz:
@LinusBorg @zigomir Thoughts?

LinusBorg:
Looks fine.

Additionally (but not nessessarily in this PR), we could replace the render function in main.js with a template?

chrisvfritz:
@LinusBorg Yeah, I like that idea a lot!

PR #326, opened 2016-10-18
add the webpack-md5-hash to optimise the chunkhash

luxp:
Add the webpack-md5-hash plugin to prevent the js chunkhash being updated when only the css file is being updated.


chrisvfritz:
While I like the idea, it looks the plugin [isn't very reliable](https://github.com/erm0l0v/webpack-md5-hash/issues/5). I think not invalidating often enough would actually be worse than invalidating too often.

PR #327, opened 2016-10-18
Use a template instead of a render function for users of the standalone template

chrisvfritz:
As @LinusBorg brought up in a separate PR, this might be a good idea to prevent those (typically less advanced) users from having to learn about render functions right away. Thoughts?


LinusBorg:
LGTM

Issue #325, opened 2016-10-18
babel-runtime not installed by default prevent webpack to build the bundle properly (ie. when using the 'typeof' keywork in the source code) when using an old version of npm (v1.4.21)


AlexandreBonneau:
I created a new project with the following command line : `vue init webpack my-project-name`.

Everything looked fine and webpack could compile my ES6 code without problem.

However, as soon as I added the `typeof` keyword in my code, or used an external library (ie. d3.js), webpack/babel output lots of errors like :
```
» npm run dev
 
> cdz@1.0.0 dev /home/user/dev/vuejs
> node build/dev-server.js
 
Listening at http://localhost:8080
 
webpack: wait until bundle finished: /index.html
webpack: wait until bundle finished: /__webpack_hmr
webpack: wait until bundle finished: /__webpack_hmr
webpack: wait until bundle finished: /index.html
webpack built a6f3de609bff15fc99ae in 36500ms
Hash: a6f3de609bff15fc99ae
Version: webpack 1.13.2
Time: 36500ms
Asset     Size  Chunks       Chunk Names
app.js  3.24 MB       0       app
 
ERROR in ./src/js/utils.js
Module not found: Error: Cannot resolve module 'babel-runtime/helpers/slicedToArray' in /home/user/dev/vuejs/src/js
@ ./src/js/utils.js 1:296-342
 
ERROR in ./src/js/utils.js
Module not found: Error: Cannot resolve module 'babel-runtime/helpers/typeof' in /home/user/dev/vuejs/src/js
@ ./src/js/utils.js 1:416-455
 
ERROR in ./src/js/utils.js
Module not found: Error: Cannot resolve module 'babel-runtime/core-js/array/from' in /home/user/dev/vuejs/src/js
@ ./src/js/utils.js 1:512-555
 
ERROR in ./src/js/utils.js
Module not found: Error: Cannot resolve module 'babel-runtime/core-js/set' in /home/user/dev/vuejs/src/js
@ ./src/js/utils.js 1:606-642
 
ERROR in ./src/lib/d3/d3.js
Module not found: Error: Cannot resolve module 'babel-runtime/core-js/math/sign' in /home/user/dev/vuejs/src/lib/d3
@ ./src/lib/d3/d3.js 1:23-65
 
ERROR in ./src/lib/d3/d3.js
Module not found: Error: Cannot resolve module 'babel-runtime/core-js/math/log2' in /home/user/dev/vuejs/src/lib/d3
[...]

PR #330, opened 2016-10-23
automatically check versions of node and npm

chrisvfritz:
@zigomir Review plz? 🙂

### DIFF
--- a/template/build/build.js
+++ b/template/build/build.js
@@ -1,4 +1,5 @@
 // https://github.com/shelljs/shelljs
+require('./check-versions')()
 require('shelljs/global')
 env.NODE_ENV = 'production'
 
--- a/template/build/check-versions.js
+++ b/template/build/check-versions.js
@@ -1,0 +1,45 @@
+var semver = require('semver')
+var chalk = require('chalk')
+var packageConfig = require('../package.json')
+var exec = function (cmd) {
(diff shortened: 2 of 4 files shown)


zigomir:
Hi @chrisvfritz! I'll do this tomorrow if it's ok with you? I'm barely from my vacation and had 100 other things to do first.

chrisvfritz:
Oh, no problem! Please take your time - it's definitely not urgent. 😄  And sorry, forgot you were on vacation. 😬

chrisvfritz:
Btw, mostly for GitHub's record, this fixes #325.

zigomir:
Eh, my fault, actually I forgot to mention anywhere (besides job of course) that I'll having one.

First thing that comest to mind is, what about `yarn`? :) But probably we don't need to care about it just yet. 

Another, more important question that comest to my mind: should we move the logic directly to `vue-cli` and inform user after download? I know user can use/install `vue-cli` with different node version, but still, this way all templates could benefit from just adding `engines` part to `package.json`.

Otherwise LGTM (I tried it locally and works fine). If it solves problems I'd merge this and try it out here and maybe later port to `vue-cli`, if or when it turns out to be useful.

chrisvfritz:
Sounds good to me! Thanks for the review. 🙂

PR #340, opened 2016-10-27
Sass router vuex

hessnd:
Added sass, removed vue boilerplate page, setup vuex store folders


pksunkara:
Apparently, the maintainers want to keep this template simple without the router and store. #296

PR #348, opened 2016-11-04
Add meta viewport to index.html

anishkny:


chrisvfritz:
There unfortunately isn't a one-size-fits-all viewport setting for all apps. If the app isn't responsive for example, this would be an undesirable. I'm honestly not sure if most users of this template are building responsive apps for not, so I'll leave this open for now in case other @vuejs-templates/collaborators have thoughts.

zigomir:
@chrisvfritz I agree with you. What we might do instead is to put a comment around viewport and let users uncomment it if they need it. What do you both think about this? cc @anishkny

yyx990803:
Closing due to inactivity. I think it's better to leave this up to users to decide.

mezka:
I'm not a collaborator but today I lost more than a couple of hours "debugging" my media queries because of this omission.

PR #354, opened 2016-11-07
Moved some devDependencies to dependencies to fix npm i --production

callumacrae:
Currently, `npm install --production && npm run build` throws an error because it doesn't have the required dependencies. I'm not sure what the system for dependencies vs devDependencies was before, but it doesn't seem that helpful - if it's already built, you don't need the vue dependency, and if it isn't build yet, you need the build dependencies. Seems to be set up to have the dependencies of a library, which it isn't - it's an app template.

This PR moves all the dependencies required for building for production to the `dependencies`.

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -13,9 +13,6 @@
     "lint": "eslint --ext .js,.vue src{{#unit}} test/unit/specs{{/unit}}{{#e2e}} test/e2e/specs{{/e2e}}"{{/lint}}
   },
   "dependencies": {
-    "vue": "^2.0.1"
-  },
-  "devDependencies": {
     "autoprefixer": "^6.4.0",
     "babel-core": "^6.0.0",
     {{#lint}}
@@ -25,9 +22,6 @@
     "babel-plugin-transform-runtime": "^6.0.0",
     "babel-preset-es2015": "^6.0.0",
     "babel-preset-stage-2": "^6.0.0",
-    "babel-register": "^6.0.0",
-    "chalk": "^1.1.3",
-    "connect-history-api-fallback": "^1.1.0",
     "css-loader": "^0.25.0",
     {{#lint}}
     "eslint": "^3.7.1",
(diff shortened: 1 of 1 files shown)


yyx990803:
`dependencies` means you use it only in your application code.
`devDependencies` means you use it in build scripts.

It's true that the distinction isn't really that useful - technically, everything is devDependency because the final build output is static files. That said, this PR doesn't seem to help with that either because `npm install --production` is a Node server only concept.

callumacrae:
What do you mean by "`npm install --production` is a Node server only concept"?

LinusBorg:
I guess what Evan means is that there is no real usecase for `npm install production` when using it with frontend project like project like this one:

* If you clone this repo, you chances are you want all dependencies because you intend to work on this project.
* If you install this project as a depenency from e.g. NPM registry, npm will only install `dependencies` anyway. And since all the code is bundled, the best thing would be none of the (dev)dependencies would be installed.

So the smartest thing would be to leave everything in devDependencies, so that it's not installed in the latter case.

callumacrae:
When deploying code written from this template, unless you have a separate build server, you want the dependencies requires to build the code, and usually not the dependencies required to test the code. It's not exactly an uncommon use case—how are you guys deploying your code?

LinusBorg:
I'll re-open this for discussion for now. Still not convinced.

Wouldn't it be preferable to add the tested build to the repo and simply pull from there instead of risking build failures on the production server?

callumacrae:
That's generally not the approach I would take, because then you can't have environmental variables. This template doesn't have them, but it's something I've added to every Vue site I've written: it's super useful for, for example, using one API endpoint on a staging server and another on the production server.

My workflow is generally:

- Run build and tests locally after writing code
- Commit code and push to github in a PR
- CI runs the build, and then runs the tests - don't merge if failing. Because the build is also being ran, this tests with updated dependencies, which is good
- Once merged into master, run tests again before deploying
- If tests pass, deploy code to staging
- Build runs on staging
- Once a week, deploy code to production, build running there

Running the build locally, while entirely valid, would restrict what I can change between the four environments I'm deploying to. The config/*.env.js files fall apart when you have more than dev/prod. There's even cases where you'd want multiple different production environments with different production environments with different variables: say, you're deploying a whitelabelled app in different environments, or you're hitting a different API for different locations.

----

Are there any disadvantages to making this change? Are many people using this template for libraries?

padcom:
The Heroku case is very interesting. Normally what I would do in such a case is I would modify my project according to Heroku-specific requirements. However it'd be great to have that information somewhere (maybe the Readme file is a good place for it?).

callumacrae:
That's a specific example of a very large number of use cases. The most common deploy method on any cloud provider or provisioned hosting, in fact.

MillerRen:
@callumacrae If other packages depends on this package, it will install all dependencies of this package very time.

callumacrae:
@MillerRen Of course. But is this template for libraries or applications? It seems more for applications, to me, and the case you're describing is for libraries.

myst729:
@callumacrae 

Agreed, the truth is that **dependencies** and **devDependencies** are not so clearly different in NPM. I mean, yes, they are different concepts, and behave differently in a way.

In the past we use bower, today everybody laughs at bower. Where's the successor? I don't find one. NPM is trying to take over, but honestly, not enough.

reverland:
Maybe `--production` can provide a way, that you can split libraries you really need to build, and others that only useful when you start a webpack dev server?

I find this useful when I start a CI that help us to build but not wanna install hot-reload or something else.

Dependencies and devDependencies really confusing here. As Evan said, `dependencies means you use it only in your application code.`. But something like webpack runtime and babel runtime may really be built into your application code. and do them should be in devDependencies?

The origin concept for deps/devdeps may not be useful, but I'm not sure, maybe we can use --production may be useful when using ci to reduce packages installed...

So, I still think this PR useful.

callumacrae:
I mean, it's not just heroku: putting dependencies required to deploy the app in `dependencies` and dependencies required to develop the app in `devDependencies` is pretty common.

yyx990803:
I still hold my original opinion (also elaborated by @LinusBorg ) - if you want to build or test, you should always do a full `npm install`. FYI, you can simply overwrite `NODE_ENV` with `npm install --production=false`.

PR #357, opened 2016-11-09
Fix typo of OccurrenceOrderPlugin

iamken1204:


template/build/webpack.prod.conf.js
```diff
@@ -36,7 +36,7 @@ var webpackConfig = merge(baseWebpackConfig, {
         warnings: false
       }
     }),
-    new webpack.optimize.OccurenceOrderPlugin(),
+    new webpack.optimize.OccurrenceOrderPlugin(),
```
LinusBorg: Fun fact: This typo originated in webpack itself in 1.0.0, and was fixed in Febuary of 2014:

https://github.com/webpack/webpack/commit/48aad3eeee3424e9150cce8df92cb5467213f40c#diff-da313f802aa85bbb07211a9399e04f95R63

Both spellings works for legacy reasons.

PR #365, opened 2016-11-15
speed up

MillerRen:
ignore watch node_modules directory


LinusBorg:
Seems useful! Is the improvement noticable?

I found other issues like this one that actually saw slowdowns: https://github.com/webpack/webpack/issues/979

MillerRen:
847ms -> 256ms in my local machine. Linux has watch files number limit.

LinusBorg:
Alright, I will test this on a project and will report back here.

yyx990803:
I tested on the latest branch with Webpack 2 and other projects using Wepback 2, and it seems there's no difference at all after adding this plugin... but thanks for the PR!

PR #372, opened 2016-11-18
Correct syntax for copying files from static directory

jingman:


LinusBorg:
You seem to have mixed a lot of other changes into this PR, like an update to webpack 2 etc. 

We can't merge this in this state.

npm-debug.log
```diff
@@ -0,0 +1,27 @@
+0 info it worked if it ends with ok
```
LinusBorg: this file belong in this PR

template/.eslintignore
```diff
@@ -1,2 +0,0 @@
-build/*.js
```
LinusBorg: Why?

template/.eslintrc.js
```diff
@@ -1,46 +0,0 @@
-module.exports = {
```
LinusBorg: ...and this shouldn't be deleted either.

There are many other changes that don'T belong here, and I will not comment on all of them.

jingman:
Woah my bad, I've been iterating on this myself in the month since I opened the PR, and I wasn't' aware it was spamming you with updates. Sorry to waste your time - thank you for your work on Vue.

LinusBorg:
it's all good. If you still can reproduce the original fix, even as a new PR, I'm, happy to merge it.

PR #381, opened 2016-11-25
relative urls to images in css point to the wrong place

taoliujun:
The question above:
https://github.com/webpack/extract-text-webpack-plugin/issues/27

And extract-text-webpack-plugin 2.0 has fixed it above:
https://github.com/webpack/extract-text-webpack-plugin/commit/a95ce1e4c414a06519f2f7f402b31cad55046d8a


LinusBorg:
Sorry, we won't update to webpack to yet.

PR #398, opened 2016-12-05
update book

pin3da:
Some links were fixed here 3225018bc145 but the documentation is still out of date.


LinusBorg:
I think instead of manually patching the generated HTML, we should update the docs by genrating new files.

/ping @chrisvfritz

pin3da:
@LinusBorg I tried to generate new files but github does not allow me to raise the PR in that way.

LinusBorg:
Yeah our team has to do that.

chrisvfritz:
Just updated!

PR #399, opened 2016-12-06
Solve eslint warnings; use webpack alias.

myst729:
- Solve eslint warnings for double quotes in nightwatch config file.
- Use webpack alias `src` over relative path in unit test entry file.


LinusBorg:
removing double quotes is fine, but why add quotes to keys that don't need any? which eslint rule/preset does require this?

myst729:
Oops, my bad. It was a batch find-replace, updated to keep only necessary quotes.

LinusBorg:
Thanks!

PR #406, opened 2016-12-11
use HashedModuleIdsPlugin to keep module ids consistent

lmk123:
As https://github.com/webpack/webpack/issues/1315#issuecomment-158677302 mentioned:

> The OccurenceOrderPlugin orders the modules in a way that most referenced modules come first and the order is the same for each compilation. This way the bundle size is as small as possible. If you change your code in a way that adds or removes modules it will affect all module ids.

I suggest use [HashedModuleIdsPlugin](https://github.com/webpack/webpack/blob/master/lib/HashedModuleIdsPlugin.js) (which introduce on https://github.com/webpack/webpack/issues/1315#issuecomment-171902187) to keep module ids consistent between compilations and not be affected by adds or removes modules, so hash will be more stable.

Related comments:

 - https://github.com/webpack/webpack/issues/1315#issuecomment-155102901
 - https://github.com/webpack/webpack/issues/1315#issuecomment-171896778
 - https://github.com/webpack/webpack/issues/1315#issuecomment-232391335

### DIFF
--- a/template/build/HashedModuleIdsPlugin.js
+++ b/template/build/HashedModuleIdsPlugin.js
@@ -1,0 +1,35 @@
+// from https://github.com/webpack/webpack/blob/master/lib/HashedModuleIdsPlugin.js
+/*
+	MIT License http://www.opensource.org/licenses/mit-license.php
+	Author Tobias Koppers @sokra
+*/
+function HashedModuleIdsPlugin(options) {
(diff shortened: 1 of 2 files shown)


LinusBorg:
closing because it's deprecated, webpack has this as a real plugin now.

PR #414, opened 2016-12-15
Fix dev-server.js to use dev port

colemickens:


colemickens:
Maybe not. I don't really understand how this config is laid out.

PR #415, opened 2016-12-15
Add an option for vue-router

crswll:
A lot of people ask how to set up the router in gitter.im so I figured an option here would be nice.


gkatsanos:
It would be very nice, I m just starting with vue-router and there's nothing useful in the vue-router documention regarding our webpack setup.

template/src/router/index.js
```diff
@@ -0,0 +1,17 @@
+import Vue from 'vue'{{#if_eq lintConfig "airbnb"}};{{/if_eq}}
+import Router from 'vue-router'{{#if_eq lintConfig "airbnb"}};{{/if_eq}}
+
+Vue.use(Router){{#if_eq lintConfig "airbnb"}};{{/if_eq}}
+
+import Hello from '../components/Hello'{{#if_eq lintConfig "airbnb"}};{{/if_eq}}
+
+export default new Router({
+  mode: 'history',
```
pholly: Yeah probably best to leave it to Router's default

pholly:
@gkatsanos I added vue-router by following its documentation and didn't have any problems. Other than adding vue-router to `package.json` by running `npm install --save-dev vue-router` I didn't change any webpack setup.

template/src/router/index.js
crswll: Agreed. I'll remove it.

gkatsanos:
seems like this was merged after all ?..

pksunkara:
Heh. Looks like I was too early with my vue-router PR. https://github.com/vuejs-templates/webpack/pull/296

Seriously don't understand why it wasn't accepted then but is accepted now.

yyx990803:
@pksunkara apologies that I didn't pay more attention to your PR earlier - it does look very similar to this one. That said, it by no means indicate that your contribution was not welcome - it's mostly that my thoughts on what the template should include do change over time. At the time of your PR the 2.x router wasn't considered very stable but now it is - so I'm more confident in including as part of the default options now than I was a few months ago. Hope that explains it.

pksunkara:
@yyx990803 No problem. If you can, please read through [this](https://github.com/vuejs-templates/webpack/pull/296#discussion_r82073211) comment on that PR and tell me what your thoughts are about it. Thanks.

PR #416, opened 2016-12-15
babel stage update

qw3n:
Update babel stage-2 to stage 1 which includes the import/export syntax
`export * as name from 'module';`. Which is useful for writing index.js files that you want batch import and then export.


LinusBorg:
I think stage-2 is fare enough from the current standard for a default template. Whoever decides to go beyond that should do so on their own, IMHO

PR #419, opened 2016-12-22
Move es-lint code to the webpack.dev.conf.js

graph1994:
Moved the lint code from the webpack.base.conf.js to the webpack.dev.conf.js as it shouldn't be needed for production from my understanding.

### DIFF
--- a/template/build/webpack.base.conf.js
+++ b/template/build/webpack.base.conf.js
@@ -35,22 +35,6 @@ module.exports = {
     fallback: [path.join(__dirname, '../node_modules')]
   },
   module: {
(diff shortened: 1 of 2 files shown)


LinusBorg:
I don't think it's  good idea to not have eslint run when you build for production.

Yes, you *can* run it independently with `npm run lint` - but currently, building is not coupled to that. If we move the webpack linting config from base to dev, this would make it possible to build production code with linting errors. Not a possibility I would want for this template.

egoist:
um what about moving it to prod? eslint errors during development are very annoying 😂

PR #420, opened 2016-12-23
FIX: JSON.stringify env object

zzhjerry:
I recently need to define an object in  `config/env.dev.js` like this:

```
module.exports = merge(prodEnv, {
  NODE_ENV: '"development"',
  GLOBAL_VAR1: {
    PROP1: "someValue"
  }
})
```

when I was trying to load to "someValue" into my code with `process.env.GLOBAL_VAR1.PROP1`, I got `ReferenceError` for `someValue`.

Adding `JSON.stringify` solved my problem.

### DIFF
--- a/template/build/webpack.dev.conf.js
+++ b/template/build/webpack.dev.conf.js
@@ -19,7 +19,7 @@ module.exports = merge(baseWebpackConfig, {
   devtool: '#eval-source-map',
   plugins: [
     new webpack.DefinePlugin({
-      'process.env': config.dev.env
+      'process.env': JSON.stringify(config.dev.env)
     }),
     // https://github.com/glenjamin/webpack-hot-middleware#installation--usage
     new webpack.optimize.OccurrenceOrderPlugin(),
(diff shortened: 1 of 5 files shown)


myst729:
Did you noticed that `NODE_ENV: '"development"'` is already stringified?

zzhjerry:
@myst729 thanks for pointing it out. I removed the single quote from `'"development"'` and all others in `test.env` and `prod.env`.

I think it's more flexible to define value as plain string or js object than quoting everying into a string

myst729:
Then you should probably use single quotes, since in the template eslint is using **standard** plugin.

airt:
'process.env' in [test/unit/karma.conf.js#L26](https://github.com/vuejs-templates/webpack/blob/master/template/test/unit/karma.conf.js#L26)
```
  plugins: [
    new webpack.DefinePlugin({
      // also need `JSON.stringify`
      'process.env': require('../../config/test.env')
    })
  ]
```

yyx990803:
You should quote your strings or call `JSON.stringify` on the strings:

``` js
module.exports = merge(prodEnv, {
  NODE_ENV: '"development"',
  GLOBAL_VAR1: {
    PROP1: JSON.stringify("someValue")
  }
})
```

Using a string for `DefinePlugin` will result in something like this:

``` js
{NODE_ENV:"production"}.NODE_ENV !== "production"
```

And will cause minifier optimization to fail.

PR #422, opened 2016-12-23
cleanup build moving common functions to utils

miljan-aleksic:
The changes were tested, the Build process works exactly the same as before with few advantages: the code is cleaner and the build could execute different webpack builds easily (required in some projects).

### DIFF
--- a/template/build/build.js
+++ b/template/build/build.js
@@ -4,9 +4,8 @@ require('shelljs/global')
 env.NODE_ENV = 'production'
 
 var path = require('path')
+var utils = require('./utils')
 var config = require('../config')
-var ora = require('ora')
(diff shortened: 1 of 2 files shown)


LinusBorg:
I like this.

miljan-aleksic:
Why not merge it, then? :)

LinusBorg:
Because I haven't come around to test this locally. And since there are no automated tests for this, I won't merge before I (or someone else) have done that.

yyx990803:
I'm not sure if this is necessary. The build code is only needed in `build.js`, it's not reusable/reused, so I'd rather keep it where it's actually used. Thanks for the PR though!

PR #435, opened 2017-01-07
add geckodriver for running e2e tests in firefox

myst729:
The npm package is a wrap for [geckodriver binaries](https://github.com/mozilla/geckodriver/releases), very similar to chromedriver.


myst729:
Looks like the driver for firefox is built-in?

PR #437, opened 2017-01-09
Introduce bundle analyzer plugin in production build

myst729:
In regard to production optimization, sometimes we may wonder is it possible to reduce my bundle size? What library contributes the most? For such cases, a bundle analyzer report is very helpful. See [npm](https://www.npmjs.com/package/webpack-bundle-analyzer) for more about the plugin.

### DIFF
--- a/template/README.md
+++ b/template/README.md
@@ -13,6 +13,9 @@ npm run dev
 
 # build for production with minification
 npm run build
+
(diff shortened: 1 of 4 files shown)


MaxMilton:
I'd prefer to see this as an option during template install rather than the default.

PR #440, opened 2017-01-10
Use #cheap-module-eval-source-map for development

CodinCat:
The build speed of `cheap-module-eval-source-map` is faster than just `eval-source-map`, and it works very well in most scenarios.

Thus I'd recommend to use cheap-module-eval-source-map for development.

webpack docs: https://webpack.github.io/docs/configuration.html#devtool

### DIFF
--- a/template/build/webpack.dev.conf.js
+++ b/template/build/webpack.dev.conf.js
@@ -15,8 +15,8 @@ module.exports = merge(baseWebpackConfig, {
   module: {
     loaders: utils.styleLoaders({ sourceMap: config.dev.cssSourceMap })
   },
-  // eval-source-map is faster for development
-  devtool: '#eval-source-map',
+  // cheap-module-eval-source-map is faster for development
+  devtool: '#cheap-module-eval-source-map',
   plugins: [
     new webpack.DefinePlugin({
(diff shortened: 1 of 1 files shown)


MaxMilton:
Seams to work as expect on my setups :)

myst729:
I personally use `#source-map`. `eval-source-map` is fast but I found it cannot debug JS in *.vue via Chrome DevTools.

mdartic:
I agree with @myst729, I use only `#source-map`, because it allows me to debug in Firefox.

Chrome seems working well with both `eval-source-map` and `cheap-module-eval-source-map`.

@yyx990803 could we imagine that in the cli there is a question to ask the user which strategy he wants to use ?

myst729:
@mdartic 

Well, I don't think it's a good idea. The interactive wizard should focus the high level decisions. Let the detailed configurations stay in the files for manual adjustment. Otherwise the wizard will inevitably turn into a pain.

mdartic:
Yep, I understand.

yoyo837:
Please keep the following comments：cheap-module-eval-source-map is faster for development, **but chrome doesn't stop at breakpoints**.

PR #447, opened 2017-01-16
Set `env: { browser: true }` in .eslintrc.js

crswll:
This seems like a pretty sensible default so here's a PR.


F-loat:
Your changes seems will only appear in standard.

crswll:
Good eye!

PR #448, opened 2017-01-17
explicitly named loaders

faisalhakim47:
to fix this issues: http://stackoverflow.com/questions/29883534/webpack-node-modules-css-index-js-didnt-return-a-function


simplesmiler:
Also https://webpack.js.org/guides/migrating/#automatic-loader-module-name-extension-removed

LinusBorg:
Hi @faisalhakim47 

Thatnks for the PR, but I will close this in favor of #457, which includes this change amongst others necessary for webpack 2.2

PR #453, opened 2017-01-20
fix open browser always

### LINKED ISSUE #425: everytime I ran npm run dev, it opens browser
I don't need `unit e2e` so I did not select them when initializing the template.
But now every time it opens browser but I just rerun `npm run dev`, that's really annoying

I think there may be some wrong logic in dev-server:

``` js
var webpackConfig = {{#if_or unit e2e}}process.env.NODE_ENV === 'testing'
  ? require('./webpack.prod.conf')
  : {{/if_or}}require('./webpack.dev.conf')
```

``` js
if (process.env.NODE_ENV !== 'testing') {
    opn(uri)
  }
```

or if there is any option like disable-opening-browser?

F-loat:
fix #425

### DIFF
--- a/template/build/dev-server.js
+++ b/template/build/dev-server.js
@@ -12,6 +12,7 @@ var webpackConfig = {{#if_or unit e2e}}process.env.NODE_ENV === 'testing'
 
 // default port where dev server listens for incoming traffic
 var port = process.env.PORT || config.dev.port
+var autoOpenBrowser = config.dev.autoOpenBrowser || true
 // Define HTTP proxies to your custom API backend
 // https://github.com/chimurai/http-proxy-middleware
 var proxyTable = config.dev.proxyTable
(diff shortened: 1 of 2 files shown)


template/build/dev-server.js
```diff
@@ -12,6 +12,7 @@ var webpackConfig = {{#if_or unit e2e}}process.env.NODE_ENV === 'testing'
 
 // default port where dev server listens for incoming traffic
 var port = process.env.PORT || config.dev.port
+var autoOpenBrowser = config.dev.autoOpenBrowser || true
```
chrisbraddock: If dev config is: `autoOpenBrowser: false` this evaluates to true

F-loat: Sorry, my mistake. I mean set default to `true`, but seems set to `false` is more concise. I will pull a new request and close this because the repository is deleted.

chrisbraddock: I went with `var autoOpenBrowser = config.dev.autoOpenBrowser === false`. Thanks for attempting to get this in. It's the first thing I toggle off. :)

F-loat: Not at all. :)

PR #461, opened 2017-01-23
Is exclude really necessary?

gurghet:
I mean, it’s hindering proper import of .vue components.

### DIFF
--- a/template/build/webpack.base.conf.js
+++ b/template/build/webpack.base.conf.js
@@ -50,8 +50,7 @@ module.exports = {
         loader: 'eslint',
         include: [
           path.join(projectRoot, 'src')
-        ],
-        exclude: /node_modules/
+        ]
       }
     ],
     {{/lint}}


msokk:
It is a common theme to blacklist `node_modules` in different loaders such as babel, eslint,etc. Otherwise webpack would have to process tons of files under `node_modules`.

Your current modification would allow eslint-loader to lint all files under node_modules.

You can use `include` to explicitly whitelist stuff.

gurghet:
> Otherwise webpack would have to process tons of files under `node_modules`.
I tried and it does not process the `node_modules`

msokk:
Sorry, I misunderstood you without the context, I didn't notice the include variable. So exclude is not needed. Hopefully that Webpack 2 PR will change this.

yyx990803:
Removed in latest webpack 2 updates. Thanks for the PR though!

PR #473, opened 2017-02-01
Adding #210: HTML linting (replacing #224)

ianwalter:
@chrisvfritz Added HTML linting as a new rule instead of including it with ESLint.


ianwalter:
Apologies, temporarily closing until I can fix an issue.

PR #479, opened 2017-02-04
fix require eslint-friendly-formatter without decision

F-loat:
Don't use eslint will be dished out the "Cannot find module" error.


kazupon:
Fixed with https://github.com/vuejs-templates/webpack/commit/0b8051f9b60270cd6883ad39f643da977d050e06

PR #484, opened 2017-02-04
Compress css file

limichange:


template/build/webpack.prod.conf.js
```diff
@@ -33,6 +34,10 @@ var webpackConfig = merge(baseWebpackConfig, {
         warnings: false
       }
     }),
+    new OptimizeCssAssetsPlugin({
+       assetNameRegExp: /\.css$/,
+       cssProcessorOptions: { discardComments: { removeAll: true } }
```
tomByrer: I *personally* believe that it is a best practice to leave in [`/*! important */` comments](https://github.com/twbs/bootstrap/blob/v4-dev/dist/css/bootstrap.css#L1) since that is typically where source name, version, copyright, & license is held:
* good for devs make it easier to find sources & versions if sourcecode isn't available
* lawyers & CEOs/CTOs need to know who owns what code of a project, especially when project ownership changes hands
* MIT & other OSS licenses tend to make it mandatory for "The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software."

So IMHO the default `{ discardComments: { removeAll: false } }` is better.  Hard to remember to flip over for lea

limichange: You are right. I will update it.

template/build/webpack.prod.conf.js
```diff
@@ -33,6 +34,9 @@ var webpackConfig = merge(baseWebpackConfig, {
         warnings: false
       }
     }),
+    new OptimizeCssAssetsPlugin({
+       assetNameRegExp: /\.css$/
```
KingMario: This is the default configuration, could be omitted.

limichange: get it.

MaxMilton:
I get better results using clean-css vs cssnano (which this plugin uses). So I would prefer if we used `clean-css-loader` or left it up to the user to do their own advanced optimisations.

limichange:
This PR isn't better, hope somebody can find better way.

PR #485, opened 2017-02-04
loader and fallbackLoader are deprecated

### LINKED ISSUE #480: loader and fallbackLoader are deprecated
Running e2e test triggers warnings:
```
fallbackLoader option has been deprecated - replace with "fallback"
loader option has been deprecated - replace with "use"
```

Warning goes away if [build/utils.js](https://github.com/vuejs-templates/webpack/blob/master/template/build/utils.js#L31) is updated replace
```js
      return ExtractTextPlugin.extract({
        loader: sourceLoader,
        fallbackLoader: 'vue-style-loader'
      })

```
with
```js
      return ExtractTextPlugin.extract({
        use: sourceLoader,
        fallback: 'vue-style-loader'
      })

```

limichange:
Fixed #480

### DIFF
--- a/template/build/utils.js
+++ b/template/build/utils.js
@@ -29,8 +29,8 @@ exports.cssLoaders = function (options) {
     // (which is the case during production build)
     if (options.extract) {
       return ExtractTextPlugin.extract({
-        loader: sourceLoader,
-        fallbackLoader: 'vue-style-loader'
+        use: sourceLoader,
+        fallback: 'vue-style-loader'
       })
     } else {
       return ['vue-style-loader', sourceLoader].join('!')
--- a/template/package.json
+++ b/template/package.json
@@ -48,7 +48,7 @@
     {{/lint}}
     "eventsource-polyfill": "^0.9.6",
     "express": "^4.14.1",
-    "extract-text-webpack-plugin": "^2.0.0-rc.2",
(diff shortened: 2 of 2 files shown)


amirrustam:
Can confirm the fix works.

limichange:
@LinusBorg Can you merge this ? many people has this problem.

alfeg:
I have to add one more change - to make webpack 2 happy
```javascript
    function generateLoaders(loaders) {
        var sourceLoader = loaders.map(function (loader) {
            var extraParamChar
            if (/\?/.test(loader)) {
                loader = loader.replace(/\?/, '-loader?')
                extraParamChar = '&'
            } else {
                loader = loader + '-loader'
                extraParamChar = '?'
            }
            return loader + (options.sourceMap ? extraParamChar + 'sourceMap' : '')
        })

        // Extract CSS when that option is specified
        // (which is the case during production build)
        if (options.extract) {
            return ExtractTextPlugin.extract({
                use: sourceLoader,
                fallback: 'vue-style-loader'
            })
        } else {
            return ['vue-style-loader', sourceLoader.join('!')].join('!')
        }
    }
```
use accept arrays, not a string. So I moved `.join('!')` down the road

limichange:
🤔 
https://github.com/webpack-contrib/extract-text-webpack-plugin#extract

kazupon:
Fixed with #490

PR #495, opened 2017-02-06
Cannot find module 'eslint-friendly-formatter'

UngererFabien:
When using the template without eslint the package is still require but not present in package.json


UngererFabien:
oops sorry it's a duplicate

PR #496, opened 2017-02-06
Fix a bug when not using JSLint

snapeuh:
Hello,

I created a webpack project with vue-cli and noticed that I had an error when running `npm run dev` :

```
Error: Cannot find module 'eslint-friendly-formatter'
```

I added the correct code so that the module doesn't get called when you chose to not use any JSLint tool.

Regards,

Romain.


kazupon:
Fixed with 0b8051f

PR #501, opened 2017-02-08
use babel-preset-env

graingert:
babel-preset-es2015 is getting a bit old in the tooth, babel-preset-env
is automatically updated and allows selective down-leveling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,7 @@ $ npm run dev
   - Use `npm run build --report`to build with bundle size analytics.
 
 - `npm run unit`: Unit tests run in PhantomJS with [Karma](http://karma-runner.github.io/0.13/index.html) + [Mocha](http://mochajs.org/) + [karma-webpack](https://github.com/webpack/karma-webpack).
-  - Supports ES2015 in test files.
+  - Supports ES2015+ in test files.
   - Supports all webpack loaders.
   - Easy mock injection.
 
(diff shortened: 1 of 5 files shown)


limichange:
+1

```bash
# Test
$ vue init graingert/webpack#babel-preset-env project
```

MaxMilton:
+1 

I've been using `babel-preset-env` in all my babel projects, it's great!

LinusBorg:
@graingert Thanks fpr the PR. I think we should do the switch to babele-env, but I think we can improve a couple of things further:


###  Polyfills

Add a new option to meta.json: "Include polyfills?" If yes, we should add `import 'babel-polyfill'` to main.js, which env will [transform](https://github.com/babel/babel-preset-env#usebuiltins) to only include what's required for the target browsers.

### Default Targets

Maybe we could include a default browser target definition to .babelrc so people can see that it's easy to change? maybe `["kast 4 versions", "not IE <9"]` ? Just an idea, will be thankful to have someone with a better knowledge about browser support suggest something.

graingert:
Sure but maybe not in this PR. I think it's worth it to get off ES2015 first

LinusBorg:
Thanks!

yyx990803:
Hey all - hate to say it but I have to revert this. The reason is that once you configure target browsers, some ES6 features will be left untouched, and it would break UglifyJS. The alternative is to use [babili](https://github.com/babel/babili) for minification, which is unfortunately ~3 times slower than UglifyJS, and would make builds significantly slower.

I think for now we will use `babel-preset-latest` instead.

graingert:
babili when integrated with the rest of the babel pipeline isn't that much slower

yyx990803:
@graingert you can't just apply it to your application source code - you need to minify the entire bundle, so using it via a Babel preset won't fly. Instead we'd have to use `babili-webpack-plugin`, which in my local testing results in 100% slower build.

Issue #200, opened 2016-07-20
Better publicPath in webpack.conf


zhangxin840:
Setting publicPath of webpack.conf to '/' is unnecessary. This will break the app if served form local file system, or a PhoneGap package.

chrisvfritz:
@zhangxin840 I believe that was originally added because it was required for [code splitting](http://webpack.github.io/docs/code-splitting.html). Have you checked that code splitting still works when it's not set?

mixedCase:
I'm a bit new to the whole SPA/JS/WebPack/modern webdev kerfuffle, could I be of help somehow? Building with `build.assetsPublicPath` set to `'./'` the default template (also adding the <Hello> component) works just fine and all scripts seem to be built correctly and load. Is that example too simple to test what you believe may break @chrisvfritz ?

zhangxin840:
@chrisvfritz I set assetsPublicPath to '' and nothing goes wrong(at least in the project).

LinusBorg:
`PublicPath` != `assetsPublicPath`

mixedCase:
@LinusBorg Regardless it fixes the problem. I now am able to run my application over file:// with no issues. Everything builds just fine, I can serve the files just fine.

LinusBorg:
Yes, but we are not discussing `assetPublicPath` here.

mixedCase:
@LinusBorg I think you're losing sight of the problem in favor of the details. In this case @zhangxin840 mentioned the problem explicitly. Changing assetsPublicPath to './' (instead of publicPath) solves the problem and everything seems to work on my end just fine afterwards, which was @chrisvfritz's worry.

In any case, how would I go about reproducing the problem you mentioned? Do you mean LiveReload breaks? Because the change is only applied to the build script, not dev.

LinusBorg:
when you use vue-router (or change the url history any other way), a relative publicpath (and most likely, also assetPublicPath will break your build. 

At least I personally experienced that webpack will try to fetch chunks (from code splitting) relative to the current URL if public path is not absolute - which will fail if you are not in the root.

mixedCase:
I use vue-router in my project. I have not touched publicPath, only assetPublicPath. As I said, everything builds and works just fine.

LinusBorg:
have you used code-splitting?

mixedCase:
@LinusBorg I'm using ES6 modules. They compile together just fine.

mixedCase:
[...]

PR #522, opened 2017-02-15
Opening index.html works with file://

guilhermefelipesilva:
Open index.html file without import errors (css and js) and without needing a server


asselin:
Relevant to #200

aspectis:
Unfortunately, this breaks relative CSS imports - [an issue with file-loader as well as url-loader [which seems to be unresolved](https://github.com/webpack-contrib/file-loader/issues/46).

LinusBorg:
Sorry, will have to close this PR. 

For now, the absolute path will stay, because relative paths cause vArious problems with vue-router (mostly in history mode) and lazy loading. 

We should rather introduce documentation that explains when to use absolute or relative paths and what to watch out for.

PR #530, opened 2017-02-18
Reference CSS resources using relative paths

asselin:
When referencing resources (e.g. fonts) in a CSS file, absolute paths are put into the generated CSS file.  If assetsPath is absolute (e.g. '/'), this is OK, but if it's relative (e.g. './'), production builds won't be able to find the assets (because of the absolute rather than relative path).

I created a repo to discuss this issue: https://github.com/asselin/vue-webpack-font-load-bug.  The project pulls in foundation-icons, which references a font in the CSS.  I added 1 icon under "Essential links" in Hello.vue.

Branch  `build_with_relative_assetsPath` changes the assetsPath to './', and showcases the issue.  The font file won't be found if you build for production because the URL put into the CSS is `static/fonts/foundation-icons.a188c2f.woff` (vs. `/static/fonts/foundation-icons.a188c2f.woff` if assetsPath is '/').

Branch `relative_assets_path_with_url-loader_change` applies my fix, which changes the URL in the CSS file to be `../fonts/foundation-icons.a188c2f.woff`.

Branch `Base_template_with_url-loader_change` shows that my change with assetsPath set to '/' still works for dev and prod builds.

### DIFF
--- a/template/build/webpack.base.conf.js
+++ b/template/build/webpack.base.conf.js
@@ -69,7 +69,9 @@ module.exports = {
         loader: 'url-loader',
         query: {
           limit: 10000,
-          name: utils.assetsPath('fonts/[name].[hash:7].[ext]')
+          name: '[name].[hash:7].[ext]',
+          outputPath: utils.assetsPath('fonts/'),
+          publicPath: process.env.NODE_ENV === 'production' ? '../fonts/' : utils.assetsPath('fonts/')
         }
       }
     ]


LinusBorg:
Wouldn't that break when we use vue-router in history mode? as soon as we navigate to some route, the relative path be wrong...

LinusBorg:
Related PR: #541

asselin:
@LinusBorg I can give it a test, but I don't think it will break, because the font's URL is relative to the CSS file-- assuming you can load the CSS file, the font's relative URL should be valid.

LinusBorg:
> because the font's URL is relative to the CSS file

Oh, right. Still wanna test it to be sure, webpack can be ticky with paths and stuff :D

asselin:
I tested with vue-router history mode, and it worked fine, no issues. @LinusBorg, would it be possible to merge this one please?

LinusBorg:
I'm currently setting up a test with lazy-loading routes, asking for a little more patience. thanks :)

LinusBorg:
So I tested it, and it's like I thought: CSS in lazy-loaded routes (in history mode) stops working with this change.

When we are at `/test1/subtest1`, then the `SubTest1` component injects its CSS into the head in a `<style>`tag, with tese relative paths, e.g. `../fonts` - which now points to `/test1/fonts` - which doesn't exist.

So, to summarize:

* If we leave absolute paths for assets, those don't work with relative `publicPath` in webpack if the app is served from a subdirectory.
* If we change it to relative paths, then lazy-loaded route components with vue-router produce invalid CSS paths to assets.

Neither are optimal. But if I have to choose, I would vote to stay with absolute paths (for both assets and `publicPath`). Why?

* With asolute paths, every app can be made to work in any subdirectory with a change of the `assetsPublicPAth` to match the subdirectory and a rebuild.
* But with relative paths, Someone working at an app that is working great in dev, would have to somehow switch over from relative to absolute paths as soon as he wants to use lazy-loaded routes (which is a *very* common usecase for any bigger app).

Unless we find a way to make the second scenario much more user-friendly than it is now, I think we have to stay with absolute paths.

asselin:
@LinusBorg When I tested history mode with relative paths, I didn't run into any issues, including reloading on a subroute.  Note that for dev, I left the assetsPublicPath as '/', so there was no issue there.

I haven't tried lazy loaded routes-- do you happen to have a snippet of code or a github repo handy with your test (I'd like to see it in action to see if I can come up with any solutions).  If not, no worries, I'll code something up.

Assuming it's an issue we can't get around, based on your comment, are you saying that you'd be OK with a PR that adds an additional question (e.g. Do you want to use an absolute or relative base path?) and then configures the template accordingly (webpack config, assetsPublicPath, etc.)?

Thanks!

LinusBorg:
have the repo locally, will upload it to GH tonight.

asselin:
@LinusBorg Gentle reminder, when you have time, to upload your GH repo.  Thanks!

LinusBorg:
Wanted to, but had trouble to reliably replictae the scenarios. Either way, relative paths seem to be brittle at best.

Will probably upload the repo tonight anyway - I hop you can find your way around the codebase with a few pointers in the readme.

LinusBorg:
So here's the repo: https://github.com/LinusBorg/webpack-tempate-test-releative-paths

asselin:
@LinusBorg Apologies for taking so long to get back on this one (client work!).

I just sent you a PR to the repo you sent me above.  I think the reason you were seeing an issue with reloads is because you had `assetsPublicPath` set to './' with history mode turned on.  When I set it to '/', it worked fine.  

Also note that this PR only affects the URLs for Fonts, and since fonts are specified in CSS files, their URL is relative to the CSS file, not relative to index.html.

Thanks!

PR #549, opened 2017-02-26
Adding default recommended rules from official eslint

jveljan:


yyx990803:
Thanks for the PR, but I don't really think it's a good idea to add another eslint config... It's already [getting quite hairy](https://github.com/vuejs-templates/webpack/blob/master/template/src/main.js) making sure template files adhere to different configs. I think we'll stick with the two we have for now.

PR #577, opened 2017-03-06
Update .eslintrc.js to make vuex work with airbnb eslint rules.

littlehaker:
By default, vuex mutations
```javascript
  mutations: {
    increment (state) {
      // mutate state
      state.count++
    }
  }
```
cannot pass airbnb eslint rules. 

Update .eslintrc.js to make it work.

### DIFF
--- a/template/.eslintrc.js
+++ b/template/.eslintrc.js
@@ -44,6 +44,8 @@ module.exports = {
       'js': 'never',
       'vue': 'never'
     }],
+    // allow vuex
+    'no-param-reassign': ["error", { 'props': false }],
     // allow optionalDependencies
     'import/no-extraneous-dependencies': ['error', {
       'optionalDependencies': ['test/unit/index.js']


Aladdin-ADD:
- it is not a good practice, as it disable checking the rule to all files.
- a better way is disabling rules with inline comments.

template/.eslintrc.js
```diff
@@ -44,6 +44,8 @@ module.exports = {
       'js': 'never',
       'vue': 'never'
     }],
+    // allow vuex
+    'no-param-reassign': ["error", { 'props': false }],
```
Aladdin-ADD: it is not a good practice, as it disable checking the rule to all files.
a better way is disabling rules with inline comments.

littlehaker: I agree. However, it's very unfriendly to throw errors when using official library Vuex.

pksunkara:
You can use the `ignorePropertyModificationsFor` array for `no-param-reassign` and add `state` to it instead of disabling it completely

LinusBorg:
This was solved by another PR recently. Sorry ;/

PR #578, opened 2017-03-06
add stylelint to template

nicholaslee119:
add stylelint to webpack template


grantcarthew:
Doesn't Vuejs use standard already?

nicholaslee119:
decline

PR #582, opened 2017-03-07
Update .babelrc for preset-env for test

unfw:
Preset env was changed for for dev/prod but not for unit test


LinusBorg:
fixes #583

PR #591, opened 2017-03-10
Add shims for all PhantomJS missing features

adambiggs:
Replaces `function-bind` with [`karma-phantomjs-shim`](https://github.com/tschaub/karma-phantomjs-shim) which adds shims for all PhantomJS missing features.

This fixes an issue I ran into where tests were failing in PhantomJS but passing in other browsers, because my code used `String.prototype.startsWith()`.


LinusBorg:
Thanks!

PR #603, opened 2017-03-16
Fix #598 - Split vendor css into its own file

### LINKED ISSUE #598: Split vendor css into its own file
This is a followup to [bootstrap-vue/bootstrap-vue#143 - Document including Bootstrap CSS](https://github.com/bootstrap-vue/bootstrap-vue/issues/143), in which I'm asking for advice to integrate CSS dependencies into my `vuejs-templates/webpack`-based app.

In comment https://github.com/bootstrap-vue/bootstrap-vue/issues/143#issuecomment-286252927, I'm advised to configure my webpack build to emit two separate bundles `app.css` and `vendor.css`. This for the same reasons it's [done for the js](https://github.com/vuejs-templates/webpack/blob/master/template/build/webpack.prod.conf.js#L66-L79), in order to:

1. Limit the number of network requests, while ensuring to...
2. ... make the slowly-changing `vendor.<hash>.css` as cache-friendly as possible.

I'm now looking at my webpack build configuration, which is mostly this template untouched, and I don't think this is currently baked in it.

* Right, or did I miss it? If right, would it make sense to add such a feature to the template, and are you interested in a PR doing it?
* Taking a step back to look at my original question in https://github.com/bootstrap-vue/bootstrap-vue/issues/143, am I missing better ways to bundle external CSS?

Thanks.

ronjouch:
For the same reasons it's done for js, in order to:

1. Limit the number of network requests, ...
2. ... while ensuring to make the slowly-changing `vendor.<hash>.css` as cache-friendly as possible.

Works as expected and emits a separate vendor css+sourcemap couple.

Thanks to @asafyish for the pointer!

### DIFF
--- a/template/build/webpack.prod.conf.js
+++ b/template/build/webpack.prod.conf.js
@@ -70,7 +70,7 @@ var webpackConfig = merge(baseWebpackConfig, {
         // any required modules inside node_modules are extracted to vendor
         return (
           module.resource &&
-          /\.js$/.test(module.resource) &&
+          /\.(css|js)$/.test(module.resource) &&
           module.resource.indexOf(
             path.join(__dirname, '../node_modules')
           ) === 0


LinusBorg:
On second thought: Shouldn't any and all CSS be extracted into app.css anyway in production, since we use ExtractTextPugin?

@ronjuch your thoughts? Am I missing some use case?

ronjouch:
> *"On second thought: Wouldn't any and all CSS be extracted into app.css anyway in production, since we use `ExtractTextPugin`?"*

@LinusBorg no, see below: with the patch, vendor css is indeed separated into `vendor.css`, and is no longer present in `app.css` (see sizes below). My understanding of the pipeline is young, so I will be of little help explaining the behavior.

(Partial) output of webpack for `npm run build`, before patch:

```
       static/css/app.10c6a273cc2d92109a91a9d5195cfa92.css     139 kB       1  [emitted]         app
```

Output of webpack for `npm run build`, after patch:

```bash
       static/css/app.5c407d2ab1bf358d9d1390d26f4a2fec.css     1.6 kB       1  [emitted]         app
    static/css/vendor.12f4a531022e30ef8a191af027124cfb.css     138 kB       0  [emitted]         vendor

```

LinusBorg:
Geez forgot about this. Want to test it myself before merging, but makes sense!

LinusBorg:
I'm really embarassed that this hasn't been reviewed (and hopefully merged) yet, butit just slipped my mind over and over, and I want to test it out for myself before merging, not because I don't trust you, but because I want to understand what's going on, and I'm a practical learner.

So I will definitely try to do this within the next 10 days. Feel free to ping me if I let it slip again.

LinusBorg:
@ronjouch I just re-read the issue #598 and now I'm a bit confused - should we include .vue files in the check or not? which variant has been "battle-tested"?

ronjouch:
@LinusBorg nothing has been "battle-tested" on my side, sorry, I moved on to other things. Maybe @BartCorremans (who originally suggested adding `vue` to the regex) can comment, but I'm afraid I will no longer be very helpful here.

PR #618, opened 2017-03-22
Add vuex support

thangman22:


nickmessing:
@thangman22, every once in a while somebody wants to add `vuex` support, just like me few months ago, `webpack` template is kept as simple as possible so, unfortunately, `vuex` is not going to be added.

template/package.json
```diff
@@ -14,7 +14,8 @@
   },
   "dependencies": {
     "vue": "^2.2.2"{{#router}},
```
crswll: You can end up with a trailing comma here if you don't include `vuex` or `vue-router`.

template/package.json
```diff
@@ -93,7 +94,8 @@
     "webpack": "^2.2.1",
     "webpack-dev-middleware": "^1.10.0",
     "webpack-hot-middleware": "^2.16.1",
-    "webpack-merge": "^2.6.1"
+    "webpack-merge": "^2.6.1",{{#vuex}}
```
crswll: If you don't include vuex you'll end up with a trailing comma.

thangman22: Ok I'm done

Jack85:
So when this one ll be solved finally?

viktor-ku:
Too bad contributors here are too conservative to add a goddamn `vuex`. I personally install vuex every single time I am using this template

saintplay:
> webpack template is kept as **simple** as possible

it kinda doesn't make sense having support for "simple" features like **e2e testing** and **lint** but no **vuex** support

viktor-ku:
> webpack template is kept as simple as possible

There is `webpack-simple` template for this, I guess?

PR #630, opened 2017-03-26
Update webpack.dev.conf.js

gdomaradzki:
Fixes "extract-text-webpack-plugin - loader is used without the corresponding plugin" when you run `npm run dev`

### DIFF
--- a/template/build/webpack.dev.conf.js
+++ b/template/build/webpack.dev.conf.js
@@ -4,6 +4,7 @@ var config = require('../config')
 var merge = require('webpack-merge')
 var baseWebpackConfig = require('./webpack.base.conf')
 var HtmlWebpackPlugin = require('html-webpack-plugin')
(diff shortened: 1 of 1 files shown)


LinusBorg:
I haven't come across this error, how did this happen for you? 

It's by design that css is not extracted in dev mode, and in dev mode, the extract loader should also not be applied.

gdomaradzki:
I am not entirely sure what causes the problem, but when in dev environment, running `npm run dev` used to give me "extract-text-webpack-plugin - loader is used without the corresponding plugin", also this fixed the same problem to a group of friends.

LinusBorg:
Well then we schools investigate why that happens - simply adding this plugin even though we don't want to use it in dev mode is not an appropriate solution. 

Would you be so kind to open an issue for th8s problem so we can track it? Thanks.

gdomaradzki:
Of course, will do it as soon as possible 👍

PR #641, opened 2017-03-31
fix watch&reload for sub folders under src/

fei-shen-null:
when running dev in an ubuntu16.04 docker image (files mounted from a centos host), changes to src/componets/*.vue will not trigger recompilation thus no hotreloading. Adding an include entry solved this problem

### DIFF
--- a/template/build/webpack.base.conf.js
+++ b/template/build/webpack.base.conf.js
@@ -43,7 +43,8 @@ module.exports = {
       {
         test: /\.vue$/,
         loader: 'vue-loader',
-        options: vueLoaderConfig
+        options: vueLoaderConfig,
+        include: [resolve('src'), resolve('test')]
       },
       {
         test: /\.js$/,


LinusBorg:
Also, I can't really understand how this fixed your problem.

fei-shen-null:
@LinusBorg I'm not sure how this worked either, but if I remove `include`, hotreloading will only work on app.vue, the rest .vue file in subfolders won't even recompile after been modified. I also tried to increase fs.inotify.max_user_watches, but doesn't seem to work. I think its a docker and webpack2 problem, so don't merge it. Just in case someone has the same problem with docker, try to add an include for `src` and maybe `node_modules ` too

LinusBorg:
Mabe you just need to change the watch mode of webpack-dev-middleware in dev-server.js?

See watchOptions -> Polling

https://github.com/webpack/webpack-dev-middleware

fei-shen-null:
Adding this worked as well
```
var devMiddleware = require('webpack-dev-middleware')(compiler, {
  publicPath: webpackConfig.output.publicPath,
  quiet: true,
  watchOptions: {
    aggregateTimeout: 300,
    poll: true
  }
})
```

ale-batt:
Can't this be update and merged ?

LinusBorg:
Well, polling is not a default in webpack-dev-middleware for a reason. 

I'm rather thinking about making this an option you can activate in `/config`

PR #642, opened 2017-03-31
Fix Minimization problem with z-index (fix #614)

### LINKED ISSUE #614: After built the z-index changed
### Version
latest[mybe not just]
[zIndexBug.zip](https://github.com/vuejs-templates/webpack/files/858212/zIndexBug.zip)

### Steps to reproduce
- Just make a vue component，write z-index in style
- Run command: `npm  run dev`，everything fine
- Run command: 'npm run build`，what a mess...my z-index value changed

I think the z-index value shouldn't be changed，because we may use the third party plugin like jQuery plugin which could use z-index. In this condition, the z-index calculation is unnecessary, and make mistakes.

### What is expected?
keep my z-index value
dev

### What is actually happening?
my z-index changed
built

LinusBorg:


LinusBorg:
1. no idea why tests failed.
2. Now I can't reproduce the original problem ... darn

LinusBorg:
not needed, Evan optimized this elsewhere

PR #643, opened 2017-03-31
Change tests.sh script to allow implementing more tests.

Toilal:
This is first step to add more tests in TravisCI. We could use [expect](https://linux.die.net/man/1/expect) to generate templates with answers different than "yes", and see what happens. (expect script example [here](https://github.com/Toilal/vue-webpack-template/blob/master/tests/default.expect))

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,5 @@
 node_modules
 .DS_Store
 docs/_book
+tests/*
+!tests/*.*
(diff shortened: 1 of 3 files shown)


Toilal:
Too add more tests, simply copy tests/default.sh into another name, and edit the script to generate the project and test it.

PR #650, opened 2017-04-04
Fix typo in nightwatch.conf.js

ifraixedes:
Fix typo in the URL which point to the docs commented in the nightwatch.conf.js file


LinusBorg:
Thanks!

PR #652, opened 2017-04-04
add eslint-vue-config to the list of linter options (fix #611)

### LINKED ISSUE #611: Give vue as an option for the eslint choices.
When I'm using vue-cli and the webpack template, the options I get for:

> Use ESLint to lint your code?

Are standard, airbnb, and none. So I choose none and then manually add vue:

`npm install eslint-config-vue eslint-plugin-vue --save-dev`

Why is vue not an option in the cli?

shailendher:
eslint-vue-config is also the default option now

### DIFF
--- a/docs/linter.md
+++ b/docs/linter.md
@@ -1,6 +1,6 @@
 # Linter Configuration
 
-This boilerplate uses [ESLint](http://eslint.org/) as the linter, and uses the [Standard](https://github.com/feross/standard/blob/master/RULES.md) preset with some small customizations.
+This boilerplate uses [ESLint](http://eslint.org/) as the linter, and uses the [Vue](https://github.com/vuejs/eslint-config-vue) preset.
 
 If you are not happy with the default linting rules, you have several options:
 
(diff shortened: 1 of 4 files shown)


shailendher:
linking #611

LinusBorg:
@chrisvfritz I think this makes sense, doesn't it?

chrisvfritz:
Hmm, I'm honestly not sure. I haven't personally used the official Vue config for a big client project yet - mostly just small official or semi-official Vue projects. If there are folks in the community that are using it successfully, then I think we can go for it. I believe `eslint-plugin-html` should be unnecessary with `eslint-plugin-vue` though.

shailendher:
As starting point, perhaps we can provide it as one of the options instead of default? @prograhammer your thoughts?

prograhammer:
@shailendher I've been using it successfully in a complex project in development, but the project has yet to be pushed to production. I think it makes most sense as default (this is Vue we are talking about) and if any issues come about then of course we have the other options we can use in the mean time.

iAlexanderMoon:
I left this not note on 611 but it probably should have been here:

FYI: eslint-config-vue has been deprecated use eslint-plugin-vue-libs

chrisvfritz:
Heh, so _actually_, it's true that `eslint-config-vue` has been deprecated, but it will be replaced by rules embedded directly into `eslint-plugin-vue`. At the moment, it's being reworked to include many more rules and embedded configs.

`eslint-plugin-vue-libs` is what we'll be using for official Vue repos from now on, so it's not meant for user projects.

LinusBorg:
Given Chris' explanation I will close this PR in favor of adding eslint-plugin-vue as soon as v3.0.0 is officially out.

PR #658, opened 2017-04-06
Bumping up a few versions

DanielPe05:
This will update a few versions, one of them allowing us to get rid of that warning that's been bothering a few people (myself included) so much.
```
(node:33278) DeprecationWarning: loaderUtils.parseQuery() received a non-string value which can be problematic, see https://github.com/webpack/loader-utils/issues/56
parseQuery() will be replaced with getOptions() in the next major version of loader-utils.
```


DanielPe05:
@LinusBorg I have been testing it. I'm using both .vue and .js component files and it seems to be just fine. I will keep testing, but so far so good.

DanielPe05:
@LinusBorg I'm deploying this to production on my project as it all seems to be working just fine locally. I will let you know if hell breaks loose once it gets there. But hopefully it all goes smoothly. Take it for a spin if you wish.

LinusBorg:
Ok I think we can merge this, looked fine in my test as well.

Issue #594, opened 2017-03-12
Too many configs. Where to change '/static/' to '/custom/'?


Joshfindit:
(using `npm run dev` for now)
Places I've attempted (yes, I know it has the same code quality of a scribble):

1. `build/webpack.dev.conf.js`

```js
module.exports = merge(baseWebpackConfig, {
...,
  dev: {
    assetsPublicPath: '/',
    assetsSubDirectory: 'artefacts'
  }
```

2. `build/webpack.dev.conf.js`

```js
module.exports = merge(baseWebpackConfig, {
...,
  build: {
    assetsPublicPath: '/',
    assetsSubDirectory: 'artefacts'
  }
```
 **Note: both 1 and 2 result in a failure to build.** Failure: `/root/vue-cli/node_modules/webpack/lib/webpack.js:19
		throw new WebpackOptionsValidationError(webpackOptionsValidationErrors);
`

3. `config/dev.env.js`

```js
module.exports = merge(prodEnv, {
  NODE_ENV: '"development"',
  build: {
    assetsPublicPath: '/',
    assetsSubDirectory: 'artefacts'
  }
})
```
**Build succeeds, but has no effect**

4. Etc. Tried multiple variations but the results are always the same: Either `npm run dev` fails completely, or the change has no effect.

johnathankent:
@Joshfindit 
Here is what you are looking for: http://vuejs-templates.github.io/webpack/backend.html

The variable to change for the "static" directory name is in:
```
config > index.js
```

PR #659, opened 2017-04-06
clarifying and updating a few paths in code examples (#594)

johnathankent:


LinusBorg:
Will merge later when I have a chance to redeploy docs as well.

PR #660, opened 2017-04-07
Don't require NPM, let users also use Yarn (fix#656)

Aladdin-ADD:
if no npm installed, not to check npm version


LinusBorg:
Thanks!

PR #662, opened 2017-04-08
support pug as template engine

wangwenpei:


template/.gitignore
```diff
@@ -11,3 +11,6 @@ test/unit/coverage
 test/e2e/reports
 selenium-debug.log
 {{/e2e}}
+{{#pug}}
+*.vue
```
LinusBorg: Why would we want to ignore vue files?!?

template/package.json
```diff
@@ -10,7 +10,8 @@
     "unit": "cross-env BABEL_ENV=test karma start test/unit/karma.conf.js --single-run"{{/unit}}{{#e2e}},
     "e2e": "node test/e2e/runner.js"{{/e2e}}{{#if_or unit e2e}},
     "test": "{{#unit}}npm run unit{{/unit}}{{#unit}}{{#e2e}} && {{/e2e}}{{/unit}}{{#e2e}}npm run e2e{{/e2e}}"{{/if_or}}{{#lint}},
-    "lint": "eslint --ext .js,.vue src{{#unit}} test/unit/specs{{/unit}}{{#e2e}} test/e2e/specs{{/e2e}}"{{/lint}}
+    "lint": "eslint --ext .js,.vue src{{#unit}} test/unit/specs{{/unit}}{{#e2e}} test/e2e/specs{{/e2e}}"{{/lint}}{{#pug}},
+    "pug": "pug -P -E vue -b ./src -w ./src"{{/pug}}
```
LinusBorg: Can you explain what the use of this command would be? Pug files or templates in. Vue files should. Be handled by webpack, shouldn't they?

template/.eslintrc.js
```diff
@@ -50,6 +50,10 @@ module.exports = {
     }],
     {{/if_eq}}
     // allow debugger during development
-    'no-debugger': process.env.NODE_ENV === 'production' ? 2 : 0
+    'no-debugger': process.env.NODE_ENV === 'production' ? 2 : 0{{#pug}},
+    // pug config
+    'eol-last': 0,
```
LinusBorg: Why is that necessary?

wangwenpei: because pug generate code doesn't  have  end line.

template/.gitignore
wangwenpei: In our project we use pug file as template engine.
generate code should be ignore.

I never used pug-loader. I think I shoud try it first. 

this PR may not widely enough for community and looks a bad idea,
sorry for waste your time.

wangwenpei:
close it, I should tried pug-loader first.

wangwenpei:
Hi, @LinusBorg 
  thanks your suggestion. I upgrade code. please review it.

LinusBorg:
You still haven't given any explanation as to why your approach here is better than simply using pug-loader.

Without such an explanation I don't see  reason to consider this feature, sorry.

wangwenpei:
Hi, @LinusBorg 

I am confused now. I changed my code use your suggest way. 
I use `<template lang="pug">` and add `templateLanguage` choice to let user select prefer template engine.

LinusBorg:
Oh, thanks!

Well, this looks much better.

I'm still a bit unsure weither we even want to give this choice in the initial setup, as commented here:

> However, I think we generally don't want to add special support for any template engine or pre processor.

/ping @vuejs-templates/collaborators WDYT?

wangwenpei:
Agree.
This feature may need be discuss. for this reason, I set default template engine is still HTML.
this PR for me is just a lift a finger. and now I already can use in my team. wish this feature is useful and can help more people.

Thanks Vue, it's awesome!

yyx990803:
I don't think we should add more questions to the scaffolding process, because there will always be more of them. I'd rather make it easy to add whatever is needed **after** scaffolding. And I'm planning a rewrite of the webpack template in the next few weeks.

PR #666, opened 2017-04-10
Upgrade url-loader to ^0.5.8

BelinChung:
When you require a module which handle by url-loader 
`example: main.js ===> require('assets/img/logo.png'))`
An error may be reported
```
Module build failed: Error: parseQuery should get a string as first argument
```
`loader-utils@0.2.16` was fixed this issue and `url-loader@0.5.8` has been upgraded `loader-utils` to `v1.1.0`


LinusBorg:
Thanks!

PR #669, opened 2017-04-10
Adding port detection to dev-server

DanielPe05:
When default port is occupied, pick another one and start the app on that port. Very useful when running multiple projects at the same time.


LinusBorg:
e2e tests need to know the port on which the devserver is running. This change would make the port "random", possibly making e2e tests fail.

PR #676, opened 2017-04-16
alias sass-loader to scss-loader for unit testing

squirmy:
Adding this alias allows vue-loader to work in injection mode when using lang="scss" in a component.

Please see https://github.com/vuejs/vue-loader/issues/724

Thanks to @rsoule3 for finding a fix.


LinusBorg:
Thanks!

Issue #474, opened 2017-02-02
"vuex requires a promise polyfill in this browser"


jaredreich:
Re-raising this as I don't think it was solved in #260 

We should just be able to install `babel-polyfill` and then:
```js
entry: {
  'babel-polyfill': 'babel-polyfill',
  app: './src/main.js'
},
```
But that's a no go.

DevanB:
The only way I could get this error to stop was by including this in the `head` of the `index.html`:

```
<!-- Script for polyfilling Promises on IE9 and 10 -->
<script src='https://cdn.polyfill.io/v2/polyfill.min.js'></script>
```

Hope there is a better solution, as I'm not pleased with this.

LinusBorg:
@jaredreich is that supposed to mean that the Babel polyfill doesnt work or that you don't have that option for some reason? 

@DevanB so including the es6-promise polyfill in main.js as explained in #260 did not work?

jaredreich:
Hey Linus thanks for the reply. Babel-polyfill is fine as I use it actively in other non-vue projects. My point was we shouldn't have to do something hacky like inserting a lone/non-stansard require/import at the top of main.js, using the webpack entry property should work. Unfortunately I don't know how this is a problem with vuex's setup, but it's definitely not a webpack or babel-polyfill issue.

jaredreich:
If anyone else runs into this issue, I was able to get `babel-polyfill` working with this syntax:
```js
entry: {
  app: [
    'babel-polyfill',
    './src/main.js'
  ]
},
```

DevanB:
@LinusBorg Yep, I added es6-promise to the top of `main.js` (like below), but still get the vuex error.

I tried putting the babel-polyfill in `webpack.base.conf.js` like below, but that didn't work either.

```
require('es6-promise').polyfill()

import Vue from 'vue'
import App from './App'
import router from './router'
import store from './store'
import { sync } from 'vuex-router-sync'
import VueHead from 'vue-head'

sync(store, router)
Vue.use(VueHead)

/* eslint-disable no-new */
new Vue({
  el: '#app',
  router,
  store,
  render: h => h(App)
})
```

```
entry: {
  app: [
    'babel-polyfill',
    './src/main.js'
  ]
},
```

LinusBorg:
The Problem is that import statements always get hoisted to the very top of the file, before *any* other code.
So while the `require` comes first in your code, it runs after the `import` for Vuex has run. 

[...]

Issue #260, opened 2016-09-13
Vuex Promise Polyfill Issue


niallobrien:
Hi all, I'm using this template with Vue v2 RC5, Vuex and VuexFire (Firebase).
When I run the dev server and try it in IE11, I get the following error in the browser console: `SCRIPT5022: [vuex] vuex requires a Promise polyfill in this browser.`
I'm not sure why this is the case as I'd assume all JS is compiled to ES5 code...?

Please advise, thanks in advance.

yyx990803:
ES6 -> ES5 transpilation only handles syntax transformation, not polyfills (these are more like runtime features). You can either use `babel-polyfill` (which includes all ES6 related polyfills) or just `es6-promise`.

niallobrien:
Thanks for clarifying @yyx990803 :+1:

DevanB:
Is there any documentation of the process to add the polyfill @yyx990803? I've checked the Webpack template docs, but didn't find anything. My webpack skills aren't strong.

yyx990803:
@DevanB polyfills are not webpack specific https://github.com/stefanpenner/es6-promise

DevanB:
Forgive me, as I'm new to dealing with this scenario. In es6-promise's documentation [here](https://github.com/stefanpenner/es6-promise#auto-polyfill), it looks as though I can simply add this: `require('es6-promise').polyfill();` to a file. Would it be ideal to add that code in my `main.js` where Vue is being instantiated, or should something like this go in `webpack.prod.conf.js` file?

Thanks for the help @yyx990803. I know you are super busy!

yyx990803:
I'd add that to the top of your entry file

DevanB:
Found one half of my problem. I was using the Prerender SPA plugin, which seems to be having the same problems: https://github.com/chrisvfritz/prerender-spa-plugin/issues/16.

After pulling that out and trying both `babel-polyfill` (and it's instructions) and `es6-promise` (and it's instructions), I was not able to get the promise polyfill to work on IE10 and below. I hope this makes it on the radar as an issue to at least research into. I will be glad to appropriately document the problem and steps I've taken to resolve it some where, if this isn't the correct place for the discussion.

Thanks @yyx990803 for the help!

atilacamurca:
Using Babel polyfill solved the problem. Here are the steps what I did:

`npm install --save-dev babel-polyfill`

[...]

PR #678, opened 2017-04-18
Add es6-promise to Karma configuration

JanCVanB:
When using Vuex, this prevents the error
```[vuex] vuex requires a Promise polyfill in this browser```
as per @adambiggs' solution here:
https://github.com/vuejs-templates/webpack/issues/474#issuecomment-279082315

The `es6-promise` package is guaranteed to be installed
as a indirect dependency of `phantomjs-prebuilt`,
but perhaps it should be added as an explicit dependency if the user
chooses to include unit testing.

### DIFF
--- a/template/test/unit/karma.conf.js
+++ b/template/test/unit/karma.conf.js
@@ -14,7 +14,10 @@ module.exports = function (config) {
     browsers: ['PhantomJS'],
     frameworks: ['mocha', 'sinon-chai', 'phantomjs-shim'],
     reporters: ['spec', 'coverage'],
-    files: ['./index.js'],
+    files: [
+      '../../node_modules/es6-promise/dist/es6-promise.auto.js',
+      './index.js'
+    ],
     preprocessors: {
       './index.js': ['webpack', 'sourcemap']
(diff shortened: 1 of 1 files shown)


LinusBorg:
I'm not sure about this.

When people use vuex, they have to add a promise polyfill in their app anyway, so adding that polyfill here might at best be unecessary and at worst conflict with their choice the made in their app code (like bluebird).

The template itself doesn't support vuex as an init option (yet), so I don't think we should add that.

JanCVanB:
I'm not super knowledgable when it comes to webpack and ES6 stuff, so I'll take your word for it! 😄 

I'm a little confused how I successfully used Vuex multiple times before encountering this problem with the polyfill, and why others seem to get tripped up by it, too (https://github.com/vuejs-templates/webpack/issues/260). Perhaps the promise polyfill should be mentioned in the Vuex docs?

Thanks! Feel free to close this PR.

LinusBorg:
We could certainly be more specific in vuex docs, feel free to open an issue for this in vuex' repository.

Issue #667, opened 2017-04-10
Have simple JS object for configuration


ifraixedes:
I am wondering if you don't think that's nicer to have a simple and usual JS object for the configurations rather than having to wrap the string in double quotes because that object is provided as it's to the [Webpack define plugin](https://webpack.js.org/plugins/define-plugin/), but we could provide pass the JSON representation of it before [assigning to `process.env`](https://github.com/vuejs-templates/webpack/blob/master/template/build/webpack.prod.conf.js#L32)

I had that modification in my project and it was easy to make, however I created the project when this template was for webpack v1 and last week I did the update and found that it's hard and tedious to synchronize changes if you have modifications in the template build folder, so I rolled back them to have the less changes as possible in order to have a simple upgrades in the future.

Thanks for considering.

LinusBorg:
Can you specify what exactly you mean with "the configurations":
>  a simple and usual JS object for the configurations 

It seems you are talking about environment variables(?), but that's not really clear to me.

ifraixedes:
Sorry, my fault.

I'm talking about the environment variables which are in [_config_ folder, specified with the files ended with _.env.js_](https://github.com/vuejs-templates/webpack/tree/master/template/config)

LinusBorg:
...and what does your propose change would look like? I don't really see how we would get around the need for quoting strings twice - or do you suggest to turn .env.js files into JSON files?

If so, that's not a good idea, as it would be very limiting.

ifraixedes:
No, I don't suggest to turn them into JSON.

Those environment files are read in a few limited places, we only have to make a simple change in those.

I wanted to check if there is a preference to keep them as they are, because I was thinking that you prefer that people understand that those values are injected in the web application by the _Webpack DefinePlugin_, but I see that there aren't any such preferences, so if it's fine for you I'll try to send a PR in a couple of weeks with such update and it will be more clear what changes are needed.

LinusBorg:
Sure!

LinusBorg:
I'm, working on improving config with #1178

PR #684, opened 2017-04-22
Have simple JS object for configuration proposal

ifraixedes:
Proposal for #667 

At the end, my initial idea didn't work as I expected; although, as I commented, I did such change in the build templates files of the application which I'm working on, I didn't realize that the value injected in the `process.env` were expected object (the first commits on this PR reflected those changes), because webpack v1 didn't build an invalid build and I was at the beginning of the application that I'm building, so I was not using any value from the `process.env` object inside of my application at that moment, hence I never saw any issue.

When I migrated to webpack v2 template, manually; I found an error in the web console when I was serving the build (dev and prod) and I didn't understand what I missed in such upgrade, threofre I reverted my changes in the template build files and upgraded the template to the current version, without messing arround with personal changes.

Working in this proposal, I've realized that such error was due the changes that I've done in the first commit of this PR, so I had to create a simple helper function to achieve what I expressed in #667 

I haven't squashed the commits because it provide context to this commet, if you like this and you want to merge, I can squash them if you want to.

### DIFF
--- a/template/build/utils.js
+++ b/template/build/utils.js
@@ -69,3 +69,12 @@ exports.styleLoaders = function (options) {
   }
   return output
 }
+
+exports.envVarsToDefinePlugin = function (envObj) {
+  return Object.keys(envObj).map(function (k) {
+    return [k, "'" + envObj[k] + "'"]
+  }).reduce(function (o, e) {
+    o[e[0]] = e[1]
+    return o
+  }, {})
+}
--- a/template/build/dev-server.js
+++ b/template/build/dev-server.js
@@ -2,7 +2,7 @@ require('./check-versions')()
 
 var config = require('../config')
 if (!process.env.NODE_ENV) {
-  process.env.NODE_ENV = JSON.parse(config.dev.env.NODE_ENV)
+  process.env.NODE_ENV = config.dev.env.NODE_ENV
 }
 
 var opn = require('opn')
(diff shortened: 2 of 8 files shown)


posva:
What if the user wants to pass something that is not a string?

ifraixedes:
I thought that, but my rationales is, that those values come from filenames which, due their name, look that are designed to define environment variables, so I would treat them as if they were common environment variables, hence their values are strings.

I prefer to narrow the options when developing, to avoid the chances to mess the implementation because people have the options of doing so. [I'm down for allowing to pass code fragments](https://webpack.js.org/plugins/define-plugin/) in the environment variables because for me, such logic should be implemented in the application, hence which is left? Booleans, OK, not a big deal, I can make the change to pass booleans values if that's a concern

Obviously this is a personal opinion, so if you prefer to keep that flexibility, I understand.

template/build/utils.js
```diff
@@ -69,3 +69,12 @@ exports.styleLoaders = function (options) {
   }
   return output
 }
+
+exports.envVarsToDefinePlugin = function (envObj) {
+  return Object.keys(envObj).map(function (k) {
+    return [k, "'" + envObj[k] + "'"]
+  }).reduce(function (o, e) {
+    o[e[0]] = e[1]
```
rohmanhm: This line looks not documented well.

PR #688, opened 2017-04-25
Add support for linked modules

adambiggs:
After this change I was able to develop with locally-linked packages using `$ yarn link`/`$ npm link`.

### DIFF
--- a/template/build/webpack.base.conf.js
+++ b/template/build/webpack.base.conf.js
@@ -1,10 +1,11 @@
 var path = require('path')
+var fs = require('fs')
 var utils = require('./utils')
 var config = require('../config')
 var vueLoaderConfig = require('./vue-loader.conf')
 
 function resolve (dir) {
-  return path.join(__dirname, '..', dir)
+  return fs.realpathSync(__dirname + '/' + path.join('..', dir))
 }
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


airyland:
I wonder why not use default value `true` or remain unchanged as people can just edit the config? Now with value:`false`  I have to add a tip in my components document for how to import `VUX` with webpack template: setting symlinks to `true` first. 

You may know `cnpm` is the largest npm mirror in China, after using `cnpm install`, every npm module will be a symlink.  And for `VUX` components, I need to add `babel-loader` for `node_modules/vux/src/**/*.js`, but loader will not work if  symlink  is set to false.

`npm run build` will get:
image

LinusBorg:
> I wonder why not use default value `true` or remain unchanged as people can just edit the config? Now with value:`false` 

What `value` are you talking about? where do you set this?

airyland:
@LinusBorg 

yes

https://github.com/LightmakerCanada/webpack/blob/bc864770a6e8361b509f517e4a88f7fc040d8a0f/template/build/webpack.base.conf.js#L30

MaxMilton:
@airyland that line is already removed in master.

airyland:
@MaxMilton 

I can still see that line:
https://github.com/vuejs-templates/webpack/blob/master/template/build/webpack.base.conf.js#L30

MaxMilton:
Sorry, you're right! I was looking at the wrong repo 😓

LinusBorg:
Will revert his. I already tried to revert this PR, but failed to remove the `symlinks` line. Just gimme a sec.

PR #691, opened 2017-04-25
Only use .vue files in karma and ignore the rest of files

jmanuelrosa:
Currently, if you want to do unit test and we use external files, such as html, css, pug, sass, etc ... we need to load the loaders by hand in ' webpack.test.conf.js`. Vue-loader already has all loaders needed, so if we only use the .vue files from our components, vue-loader will be in charge of processing everything (templates and styles included)

Thanks :smile:


LinusBorg:
So if I understand this correctly, this will only require `.js` and `.vue` files for coverage?

PR #693, opened 2017-04-27
Bump vue version to 2.3.0

mul14:


stenote:
I think it's better to add `vue-hot-reload-api` devDependencie

egoist:
seems it's unnecessary to update minor version? It will always install the latest 2.y.z

LinusBorg:
Closing this as version is already surpassed.

PR #705, opened 2017-05-09
[docs]Add bundle size analytics command

c01nd01r:


LinusBorg:
That's now a setting in /config...

PR #718, opened 2017-05-21
add vue-hot-reload-api to devDependencies

stiekel:


heinzdmx:
Why add this? The documentation for `vue-hot-reload` states that it is not needed for normal application usage where you use `vue-loader` is used [source](https://github.com/vuejs/vue-hot-reload-api#usage)

PR #727, opened 2017-05-25
Added scripts and updated READMEs to include instructions for buildin…

luisdlopez:
…g components for distribution through npm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 .DS_Store
 docs/_book
+.idea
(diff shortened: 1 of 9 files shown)


luisdlopez:
I added this PR to this template by mistake, however, I think it won't hurt to ask you what you guys think about this.

I've been looking for a vue.js project template that includes scripts for building components to be distributed through npm (as well as for development), and I couldn't find any obvious information on this template regarding this use case, and I would be willing to continue working on this to add it to the official template if you think the way I went about doing it is right.

luisdlopez:
This template includes a `index.js` that is used as the `main` package file.
It uses this file as the entry point to build the components you want to export by running the command `npm run build-dist`.
The generated file includes all html and css (non extracted) under the folder `publish`.

PR #734, opened 2017-06-04
Fix hot reload, if install template with Yarn

TuxBoy:
Hi, 
The hot reload does not work if we install the template via yarn, when I add the last piece of code it works, after why exactly that bug with yarn, I do not know.

Thanks


LinusBorg:
This is likely not related to yarn, but some other environment you use (dropbox /docker / vagrant ...) which interfers with webpacks default way of watchigng files. 

however, polling is not a very performant way to watch files, so we should make this optional. I just submitted a PR for that (#946 ), before finding yours.

I hope it's ok if I close yours in favour of mine, which adds a more flexible config option.

Greets, Thorsten.

TuxBoy:
Ok, thanks so much for your reply.

PR #737, opened 2017-06-07
Add CDN feature

van-nguyen:
Enhances html-webpack-plugin functionality by allowing you to specify the modules you want to externalize from node_modules in development and a CDN in production.

Basically this will allow you to greatly reduce build time when developing and improve page load performance on production.

### DIFF
--- a/template/build/webpack.dev.conf.js
+++ b/template/build/webpack.dev.conf.js
@@ -4,7 +4,8 @@ var config = require('../config')
 var merge = require('webpack-merge')
 var baseWebpackConfig = require('./webpack.base.conf')
 var HtmlWebpackPlugin = require('html-webpack-plugin')
-var FriendlyErrorsPlugin = require('friendly-errors-webpack-plugin')
+var FriendlyErrorsPlugin = require('friendly-errors-webpack-plugin'){{#cdn}}
+var WebpackCdnPlugin = require('webpack-cdn-plugin'){{/cdn}}
 
 // add hot-reload related code to entry chunks
 Object.keys(baseWebpackConfig.entry).forEach(function (name) {
@@ -29,7 +30,8 @@ module.exports = merge(baseWebpackConfig, {
       filename: 'index.html',
       template: 'index.html',
       inject: true
-    }),
+    }),{{#cdn}}
+    new WebpackCdnPlugin({modules: config.build.packages, prod: false, publicPath: '/libraries'}),{{/cdn}}
     new FriendlyErrorsPlugin()
   ]
 })
--- a/template/config/index.js
+++ b/template/config/index.js
@@ -7,7 +7,19 @@ module.exports = {
     index: path.resolve(__dirname, '../dist/index.html'),
     assetsRoot: path.resolve(__dirname, '../dist'),
     assetsSubDirectory: 'static',
-    assetsPublicPath: '/',
+    assetsPublicPath: '/',{{#cdn}}
+    packages: [
+      {
+        name: 'vue',
+        var: 'Vue',
+        path: 'dist/vue.runtime.min.js'
+      },
+      {
+        name: 'vue-router',
+        var: 'VueRouter',
+        path: 'dist/vue-router.min.js'
+      }
+    ],{{/cdn}}
     productionSourceMap: true,
     // Gzip off by default as many popular static hosts such as
     // Surge or Netlify already gzip all static assets for you.
--- a/template/package.json
+++ b/template/package.json
@@ -94,6 +94,9 @@
     "vue-style-loader": "^3.0.1",
     "vue-template-compiler": "^2.3.3",
     "webpack": "^2.6.1",
+    {{#cdn}}
+    "webpack-cdn-plugin": "^1.3.4",
+    {{/cdn}}
     "webpack-dev-middleware": "^1.10.0",
     "webpack-hot-middleware": "^2.18.0",
     "webpack-merge": "^4.1.0"
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+/.idea/
 node_modules
 .DS_Store
 docs/_book
--- a/README.md
+++ b/README.md
@@ -37,6 +37,7 @@ If port 8080 is already in use on your machine you must change the port number i
   - HTML minified with [html-minifier](https://github.com/kangax/html-minifier).
   - CSS across all components extracted into a single file and minified with [cssnano](https://github.com/ben-eb/cssnano).
   - All static assets compiled with version hashes for efficient long-term caching, and a production `index.html` is auto-generated with proper URLs to these generated assets.
+  - Ability to reference libraries from a CDN of your choice when building in production, falling back to node_modules when developing.
   - Use `npm run build --report`to build with bundle size analytics.
 
 - `npm run unit`: Unit tests run in PhantomJS with [Karma](http://karma-runner.github.io/0.13/index.html) + [Mocha](http://mochajs.org/) + [karma-webpack](https://github.com/webpack/karma-webpack).
--- a/meta.js
+++ b/meta.js
@@ -42,7 +42,8 @@ module.exports = {
     },
     "cdn": {
       "type": "confirm",
-      "message": "Use a CDN when building for production?"
+      "message": "Use a CDN when building for production?",
+      "default": false
     },
     "cdnConfig": {
       "when": "cdn",
--- a/meta.js
+++ b/meta.js
@@ -40,6 +40,27 @@ module.exports = {
         }
       ]
     },
+    "cdn": {
+      "type": "confirm",
+      "message": "Use a CDN when building for production?"
+    },
+    "cdnConfig": {
+      "when": "cdn",
+      "type": "list",
+      "message": "Pick a CDN repository",
+      "choices": [
+        {
+          "name": "unpkg.com",
+          "value": "unpkg",
+          "short": "unpkg"
+        },
+        {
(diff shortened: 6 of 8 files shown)


template/build/webpack.dev.conf.js
```diff
@@ -4,7 +4,8 @@ var config = require('../config')
 var merge = require('webpack-merge')
 var baseWebpackConfig = require('./webpack.base.conf')
 var HtmlWebpackPlugin = require('html-webpack-plugin')
-var FriendlyErrorsPlugin = require('friendly-errors-webpack-plugin')
+var FriendlyErrorsPlugin = require('friendly-errors-webpack-plugin'){{#cdn}}
```
Aladdin-ADD: seems it is not related to the PR(CDN feature)?

van-nguyen: Think have to do this to prevent empty lines when #cdn is false.

template/build/webpack.dev.conf.js
```diff
@@ -29,7 +30,8 @@ module.exports = merge(baseWebpackConfig, {
       filename: 'index.html',
       template: 'index.html',
       inject: true
-    }),
+    }),{{#cdn}}
+    new WebpackCdnPlugin({modules: config.build.packages, prod: false, publicPath: '/libraries'}){{/cdn}}
```
nickmessing: Don't you need a comma at the end here?

template/package.json
```diff
@@ -93,6 +93,9 @@
     "vue-style-loader": "^3.0.1",
     "vue-template-compiler": "^2.3.3",
     "webpack": "^2.6.1",
+    {{#cdn}}
+    "webpack-cdn-plugin": "^1.3.4",
```
nickmessing: Won't this end up being an empty line when `cdn` is false and two empty lines when `cdn` is true?

template/build/webpack.dev.conf.js
nickmessing: Looks like @van-nguyen is right, that's used in other files like `package.json`

template/build/webpack.dev.conf.js
```diff
@@ -29,7 +30,8 @@ module.exports = merge(baseWebpackConfig, {
       filename: 'index.html',
       template: 'index.html',
       inject: true
-    }),
+    }),{{#cdn}}
+    new WebpackCdnPlugin({modules: config.build.packages, prod: false, publicPath: '/libraries'}),{{/cdn}}
```
van-nguyen: @nickmessing acknowledged, and has been fixed. Thanks for that, good spot.

template/package.json
van-nguyen: I've tested running `vue init van-nguyen/webpack` seems there is no empty lines, maybe it's smart enough to detect these things.

template/build/webpack.dev.conf.js
Aladdin-ADD: right. eyes had deceived me. --.

template/build/webpack.dev.conf.js
van-nguyen: @nickmessing acknowledged, and has been fixed. Thanks for that, good spot.

template/package.json
van-nguyen: I've tested running vue init van-nguyen/webpack seems there is no empty lines, maybe it's smart enough to detect these things.

van-nguyen:
@Aladdin-ADD @nickmessing is this good enough to merge? Thanks.

posva:
Mmmh, I wonder if it's a good idea to use a CDN when bundling your app with webpack because you may want to put vue+vue-router+vuex together into the same file.
I also think this should be disabled by default (haven't tested yet)

van-nguyen:
@posva What's the main advantage of having them in the same file besides hacking for performance? Specially when http2 is on the rise.

posva:
what do you mean by hacking for performance?
I'm actually asking myself the same question for CDN 😆 Maybe it's starting to be a good time to start using them again

van-nguyen:
I mean files are separate by nature and should be called as such. Putting them together and serving them in the same file is a "hack" just because of the limitations of http. When you use http2 for the better this actually hurts and you will need to "unhack" everything.

posva:
Yeah, to a certain degree:

- There is still a protocol overhead for each request compared to a single concatenated file.
- The compression of the single large file is better than many small files.
- Servers are slower serving many small files than a single large file.

(from https://medium.com/webpack/webpack-http-2-7083ec3f3ce6)

van-nguyen:
While I agree on most parts but by using a CDN you are contributing to the network of caches and in best case scenario your users won't even need to download the file (leaving gzip irrelevant) while helping other sites that uses the same CDN achieve the same. To me this is huge savings, but I'm happy to leave this off by defaut. happy to discuss.

nickmessing:
I totally agree with @van-nguyen, that was why CDNs were popular in the first place. If this feature is enabled by default - potentially in long term most websites powered by Vue are going to share same cache making both gzip and file sizes irrelevant.

santagu:
+1 
I also need this. In my company many applications will share libraries like vue-material, vuejs, vue-router (for example) and I do not want to repeat this code in each application.
Thanks for this good job, @van-nguyen

van-nguyen:
@posva I have made the option to default to `false`, can this please be reconsidered?

posva:
I'll share with the team

template/config/index.js
```diff
@@ -7,7 +7,19 @@ module.exports = {
...
     assetsSubDirectory: 'static',
-    assetsPublicPath: '/',
+    assetsPublicPath: '/',{{#cdn}}
+    packages: [
+      {
+        name: 'vue',
+        var: 'Vue',
+        path: 'dist/vue.runtime.min.js'
+      },
+      {
```
posva: this should be added only if vue-router is selected though.
I think we could do something smarter that checks if the packages `vue-router` and `vuex` are present in the package.json and build the `packages` array at runtime

van-nguyen: Good catch, acknowledged. I couldn't find any reference to `vuex`, however I have fixed for vue-router. Hope it's all good.

template/config/index.js
```diff
@@ -7,7 +7,19 @@ module.exports = {
...
     assetsRoot: path.resolve(__dirname, '../dist'),
     assetsSubDirectory: 'static',
-    assetsPublicPath: '/',
+    assetsPublicPath: '/',{{#cdn}}
+    packages: [
+      {
+        name: 'vue',
+        var: 'Vue',
+        path: 'dist/vue.runtime.min.js'
+      }{{#router}},
```
posva: Can you make it dynamic instead? checking if vue-router and vuex are in the package.json file?

van-nguyen: I'm confused, thought the `{{#router}}` switch is meant to do that? Are you suggesting to actually read the packages.json file directly and not by flag?

ta-web-paris: yes

posva: yes

template/config/index.js
```diff
@@ -1,13 +1,30 @@
...
 var path = require('path')
+{{#cdn}}
+var dependencies = require('../package.json').dependencies;
+var packages = [{
+  name: 'vue',
+  var: 'Vue',
+  path: 'dist/vue.runtime.min.js'
+}]
+
+if (dependencies['vue-router'] || dependencies.vuex) {
```
van-nguyen: @posva is this what you meant?

posva: yes, but here you're adding vue-router if vuex is in the dependencies. It could be something like:

```js
var cdnPackages = [{
    name: 'vue-router',
    var: 'VueRouter',
    path: 'dist/vue-router.min.js'
}, {
    name: 'vuex',
    var: 'Vuex',
    path: 'dist/vuex.min.js'
}]

packages = packages.concat(cdnPackages.filter(({ name }) => name in dependencies))
```

van-nguyen:
But vuex is not even part of the template?

posva:
If a user adds vuex while using the CDN feature, he'd expect to use the CDN version as well

van-nguyen:
The plugin is not smart enough to do that.. they will have to add it manually. For `vue-router` it's already part of the template, we can do. As we don't know what the user will add.. it can be anything right? Cannot cater for everything.

posva:
What plugin? 😅  I literally wrote the code for you, I didn't test it but it should be working
The reason to include vuex is that it's also an official Vue plugin

van-nguyen:
The plugin is called `WebpackCdnPlugin`, this PR is mainly for the said plugin. vuex flag does not exist yet? it can be in another PR when it does?

posva:
Oh, but it's unrelated to the plugin, we're passing on the packages ourselves. Even if vuex is not part of the template we can make the experience much better by providing support for it (since it's an official Vue plugin)

```js
var cdnPackages = [{
    name: 'vue-router',
    var: 'VueRouter',
    path: 'dist/vue-router.min.js'
}, {
    name: 'vuex',
    var: 'Vuex',
    path: 'dist/vuex.min.js'
}]

packages = packages.concat(cdnPackages.filter(({ name }) => name in dependencies))
```

van-nguyen:
It is related.. the plugin is the only one consuming the array.. and produce the cdn artefacts.. The purpose of this array is providing options for the said plugin.. nothing else..

posva:
no, building the array is unrelated to the plugin. How the array looks like **is**. I even wrote you the code...

van-nguyen:
@posva sure, hope it's all good.

PR #743, opened 2017-06-12
Add PurifyCSS Plugin for webpack and description to README

asmyshlyaev177:
Tested with bulma.io CSS framework and small todo app. Can be usefull.

### DIFF
--- a/template/build/webpack.prod.conf.js
+++ b/template/build/webpack.prod.conf.js
@@ -9,7 +9,6 @@ var CopyWebpackPlugin = require('copy-webpack-plugin')
 var HtmlWebpackPlugin = require('html-webpack-plugin')
 var ExtractTextPlugin = require('extract-text-webpack-plugin')
 var OptimizeCSSPlugin = require('optimize-css-assets-webpack-plugin')
-var PurifyCSSPlugin = require('purifycss-webpack')
 
 var env = {{#if_or unit e2e}}process.env.NODE_ENV === 'testing'
   ? require('../config/test.env')
@@ -42,22 +41,6 @@ var webpackConfig = merge(baseWebpackConfig, {
     // extract css into its own file
     new ExtractTextPlugin({
       filename: utils.assetsPath('css/[name].[contenthash].css')
-    }),
-  // Analyzes your code and remove unused CSS from resulted file
-	new PurifyCSSPlugin({
-        purifyOptions: {
-            whitelist: ['someclass']
-        },
-      // Logs out verbose logs.
-       verbose: true,
-       // An array of file extensions for determining used classes within node_modules
-       moduleExtensions: ['.vue'],
-       // Give paths to parse for rules. These should be absolute!
-      paths: glob.sync([
-        // add here more file those can contains css classes eg *.+(html|vue|jsx)')
-        path.join(__dirname, '../**/*.+(html|vue)'),
-        // other path for HTML or vue files
-      ])
     }),
     // Compress extracted CSS. We are using this plugin so that possible
     // duplicated CSS from different components can be deduped.
@@ -115,6 +98,28 @@ var webpackConfig = merge(baseWebpackConfig, {
     ])
   ]
 })
+
+if (config.build.removeUnusedCss) {
+  var PurifyCSSPlugin = require('purifycss-webpack')
+
+  // Analyzes your code and remove unused CSS from resulted file
+	new PurifyCSSPlugin({
+        purifyOptions: {
+            whitelist: ['someclass']
+        },
+      // Logs out verbose logs.
+      verbose: false,
+       // An array of file extensions for determining used classes within node_modules
+      moduleExtensions: ['.vue'],
+       // Give paths to parse for rules. These should be absolute!
+      paths: glob.sync([
+        // add here more file those can contains css classes eg *.+(html|vue|jsx)')
+        path.join(__dirname, '../**/*.+(html|vue)'),
+        // other path for HTML or vue files
+      ])
+    })
+  )
+}
 
 if (config.build.productionGzip) {
   var CompressionWebpackPlugin = require('compression-webpack-plugin')
--- a/template/build/webpack.prod.conf.js
+++ b/template/build/webpack.prod.conf.js
@@ -43,22 +43,20 @@ var webpackConfig = merge(baseWebpackConfig, {
       filename: utils.assetsPath('css/[name].[contenthash].css')
     }),
 	new PurifyCSSPlugin({
-      
         purifyOptions: {
             whitelist: ['someclass']
         },
-       
       // Logs out verbose logs.
-      verbose: false,
-      // An array of file extensions for determining used classes within node_modules
-      moduleExtensions: ['.html', '.vue'],
-      // Give paths to parse for rules. These should be absolute!
+       verbose: true,
+       // An array of file extensions for determining used classes within node_modules
+       moduleExtensions: ['.html', '.vue'],
+       // Give paths to parse for rules. These should be absolute!
       paths: glob.sync([
         path.join(__dirname, '../src/components/*.vue'),
         path.join(__dirname, '../src/*.vue'),
         path.join(__dirname, '../index.html'),
         // other path for HTML or vue files
-      ]),
+      ])
     }),
     // Compress extracted CSS. We are using this plugin so that possible
     // duplicated CSS from different components can be deduped.
--- a/template/build/webpack.prod.conf.js
+++ b/template/build/webpack.prod.conf.js
@@ -9,6 +9,7 @@ var CopyWebpackPlugin = require('copy-webpack-plugin')
 var HtmlWebpackPlugin = require('html-webpack-plugin')
 var ExtractTextPlugin = require('extract-text-webpack-plugin')
 var OptimizeCSSPlugin = require('optimize-css-assets-webpack-plugin')
+var PurifyCSSPlugin = require('purifycss-webpack')
 
 var env = {{#if_or unit e2e}}process.env.NODE_ENV === 'testing'
   ? require('../config/test.env')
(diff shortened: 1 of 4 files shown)


asmyshlyaev177:
Forget, need change 
var glob = require('glob');   
to
var glob = require('glob-all');
in webpack.prod.conf.js

template/build/webpack.prod.conf.js
```diff
@@ -41,6 +42,18 @@ var webpackConfig = merge(baseWebpackConfig, {
     new ExtractTextPlugin({
       filename: utils.assetsPath('css/[name].[contenthash].css')
     }),
+	new PurifyCSSPlugin({
+      // Give paths to parse for rules. These should be absolute! 
+        purifyOptions: {
+            whitelist: ['someclass']
```
Aladdin-ADD: what is 'someclass'? what if a user happen to naming a same classname?

asmyshlyaev177: It possible plugin accidentally will not include some css class to resulted css. So whitelist option contains name of classes those will be included forcibly.

> whitelist Array of selectors to always leave in. Ex. ['button-active', '*modal*'] this will leave any selector that includes modal in it and selectors that match button-active. (wrapping the string with *'s, leaves all selectors that include it)

More detailed description about this plugin
https://github.com/purifycss/purifycss
https://www.npmjs.com/package/purifycss-webpack

Aladdin-ADD: thanks for explanation. :)
this is just an example, and I prefer it being in a comment -- since a user may forget to change this.

asmyshlyaev177: I added more detailed description and more options, just in case.

template/build/webpack.prod.conf.js
```diff
@@ -41,6 +42,22 @@ var webpackConfig = merge(baseWebpackConfig, {
     new ExtractTextPlugin({
       filename: utils.assetsPath('css/[name].[contenthash].css')
     }),
+	new PurifyCSSPlugin({
```
MaxMilton: It's probably worth adding a comment above here to let the developer's know what this plugin is doing.

template/build/webpack.prod.conf.js
```diff
@@ -41,6 +42,22 @@ var webpackConfig = merge(baseWebpackConfig, {
...
+        purifyOptions: {
+            whitelist: ['someclass']
```
MaxMilton: Do we really need to whitelist 'someclass'?

template/build/webpack.prod.conf.js
```diff
@@ -41,6 +42,22 @@ var webpackConfig = merge(baseWebpackConfig, {
...
+        },
+      // Logs out verbose logs.
+       verbose: true,
+       // An array of file extensions for determining used classes within node_modules
+       moduleExtensions: ['.html', '.vue'],
+       // Give paths to parse for rules. These should be absolute!
+      paths: glob.sync([
+        path.join(__dirname, '../src/components/*.vue'),
```
MaxMilton: I'm not familiar with `glob-all` but will this match file in subdirectories too? Wouldn't it be better to do something like `path.join(__dirname, '../src/**/*.vue'),`?

I create directories for routing (`/routing/`) and vuex (`/store/`) so I don't think your globbing rules would match any files in those directories.

template/build/webpack.prod.conf.js
MaxMilton: Where is this coming from? `PurifyCSSPlugin` is not referenced anywhere, did you forget to import it? Does it not need packages added to `package.json` -- `purifycss-webpack` and `purify-css`?

template/build/webpack.prod.conf.js
```diff
@@ -1,4 +1,5 @@
 var path = require('path')
+var glob = require('glob-all')
```
MaxMilton: Do we not have a way to do globbing already that doesn't need an additional dependency?

template/build/webpack.prod.conf.js
asmyshlyaev177: It is example how to do it  just in case. It change nothing if user have someclass or doesn't.

template/build/webpack.prod.conf.js
asmyshlyaev177: Can make it shorter like this path.join(__dirname, '../**/*.+(html|vue)') Will match subdirectories '/routing/' too, but need match only file those can contain CSS classes.

I added it, tnx for good question.

template/build/webpack.prod.conf.js
asmyshlyaev177: Add description already.

asmyshlyaev177: My bad, fixed it.

template/build/webpack.prod.conf.js
asmyshlyaev177: Will be not obvious how to change something if go to plugin docs.

template/build/webpack.prod.conf.js
MaxMilton: You still need to add the dependencies to `package.json`...

template/build/webpack.prod.conf.js
MaxMilton: I think since you've been able to combine the glob bellow into a single pattern we no longer need `glob-all`, we can just use node's built in `glob` right? :)

asmyshlyaev177: Nope, for include few files or subdirectories need glob-all. I tried it, searched also.

asmyshlyaev177: Not sure how to do better, seems like i pushed wrong commit for pull request. 
Can i merge all these commits to one or need close this pull request and create a new one?

MaxMilton:
It's look much better now! One last thing I think is necessary is the ability to disable this functionality when not required (I use the template for non SPA apps too which wouldn't be compatible with this setup).

Could you wrap the `webpack.prod.conf.js` logic in an `if` and provide a flag in `config/index.js` to enable/disable this? It would be exactly the same as `CompressionWebpackPlugin`.

asmyshlyaev177:
> t's look much better now! One last thing I think is necessary is the ability to disable this functionality when not required (I use the template for non SPA apps too which wouldn't be compatible with this setup).

I added this in last my commit.

MaxMilton:
Just had a go at running this locally and everything works as expected.

I did have an issue where it removed some used CSS and my webpage didn't look correct. Since developers might be confused when necessary CSS is lost, should this be opt-in/disabled by default (like the `productionGzip` option)?

Anyone have thoughts on this?

Ffloriel:
My thought on Purifycss is that it's not ready yet to be implemented by default on a template.
There are some issues that are yet to be resolved [(purifycss issues)](https://github.com/purifycss/purifycss/issues).

template/build/webpack.prod.conf.js
```diff
@@ -43,22 +42,6 @@ var webpackConfig = merge(baseWebpackConfig, {
     new ExtractTextPlugin({
       filename: utils.assetsPath('css/[name].[contenthash].css')
     }),
-  // Analyzes your code and remove unused CSS from resulted file
-	new PurifyCSSPlugin({
```
bajras: Hi there, Im trying to do the similar case to remove unused CSS from Vue webpack template. I have post css and css @imports in my single component file. When I use what you have done here, i dont see its removing all the unused CSS.

asmyshlyaev177: Not all, most of it.

Ffloriel:
Purifycss is not actively maintain so I moved away from the project to have more control. I created a similar tool call purgecss.
[purgecss](https://github.com/FullHuman/purgecss) is being used in some laravel, vue, tailwindcss projects now and should be more reliable.

philwolstenholme:
There's also https://github.com/HapLifeMan/purifycss-extended which looks like it's still being worked on

LinusBorg:
I don't feel like merging this if 

* It can lead to accidentally removing used CSS
* the project doesn'T seem to be well-maintained.

I also don'tt think an opt-in is a solution, since in my opinion, people should only add this if they know what they are doing. Asking them ahead in the questionnair leads to people generating projects with this tool even though they don't know about its limitations.

focussing:
Maybe this gives another view?

[](https://forum.vuejs.org/t/purifycss-removes-styles-in-single-file-components/12123/3)

LinusBorg:
Well, I simply don't know this tool and don't plan on learning it extensively. SO to merge a PR that adds this, I would reuire extensive tests for:

* usage with normal CSS
* usage with scoped CSS enabled
* usage with CSS-modules enabled
* with and without ExtarctText

Otherwise I will be left with a template containing a tool I don't know, and not knowing what to do if the inevitable issues pop up about "CSS is wrongly left out in case X"

philwolstenholme:
For anyone that finds this issue via search, I've just implemented [purgecss](https://github.com/FullHuman/purgecss) fairly easily. 

1. Install `purgecss-webpack-plugin`: https://github.com/FullHuman/purgecss-webpack-plugin
2. Install `node-glob-all`: https://github.com/jpillora/node-glob-all
3. Require `glob-all` and `purgecss-webpack-plugin` in `build/webpack.prod.conf.js`:
```
const glob = require('glob-all')
const PurgecssPlugin = require('purgecss-webpack-plugin')
```
4. Add this to the bottom of the `plugins` section of `build/webpack.prod.conf.js`:

```
    new PurgecssPlugin({
      paths: glob.sync([
        './index.html',
        './src/*.vue',
        './src/**/*.vue'
      ])
    })
```

Ffloriel:
There is also an example with vue in the purgecss repository [here](https://github.com/FullHuman/purgecss/tree/master/examples/with-vue)

@LinusBorg I agree with the reason you gave. The tool can remove used css if it is not configured properly.  Since the webpack configuration files are still accessible after `vue-init`, I think it's better not to include that kind of tools by default. Instead, the tools should give instructions about how to use it with webpack or vue.

philwolstenholme:
If you're using Vue transitions with something like [vue2-animate](https://github.com/asika32764/vue2-animate) you'll need to add the animation class name prefixes to `whitelistPatterns`, for example:


```
      whitelistPatterns: [
        /^fade-/,
        /^slideUp-/
      ]
```

PR #744, opened 2017-06-13
make the webpack hot module reloader work with a publicPath set

scoobah36:
we have multiple apps using VUE webpack project structure and in order for each of them to be able to use the hmr they need to be deployed under each specific app context.

### DIFF
--- a/template/build/dev-client.js
+++ b/template/build/dev-client.js
@@ -1,6 +1,6 @@
 /* eslint-disable */
 require('eventsource-polyfill')
-var hotClient = require('webpack-hot-middleware/client?dynamicPublicPath=true&noInfo=true&reload=true')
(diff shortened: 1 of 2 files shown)


LinusBorg:
I haven't really understood in what context this is useful...

scoobah36:
we have multiple applications running under the same domain. in order to have them running locally all under the same port they have to be under different context. ie localhost:8080/app1, localhost:8080/app2 etc. it isn't super necessary for everyone but it don't break the existing functionality and adds new functionality so i figured i would make a pull request.



AJ Danelz


________________________________

LinusBorg:
This won't be an issue after #975 I assume. Please be so kind to open a new issue / PR if this persists after #975 has been merged

PR #745, opened 2017-06-14
fix: process cannot exit immediately when encountering build error

chikara-chan:
Such as command `npm run build && echo 'build successfully'`, when encountering build error, the console still outputs the info 'build successfully' finally.


LinusBorg:
W already merged another PR that does something similar.

Please review the changes we made -if you think those can be further improved, we welcome a nother PR from you.

PR #747, opened 2017-06-17
Enable proxyMiddleware to output more information.

thegatheringstorm:
:smile:

### DIFF
--- a/template/build/dev-server.js
+++ b/template/build/dev-server.js
@@ -47,6 +47,7 @@ Object.keys(proxyTable).forEach(function (context) {
   if (typeof options === 'string') {
     options = { target: options }
   }
+  options.logLevel = 'debug'
   app.use(proxyMiddleware(options.filter || context, options))
 })
 


MaxMilton:
I like this but I feel it should be optional because on some projects it can flood your terminal with excessive information.

template/build/dev-server.js
```diff
@@ -47,6 +47,7 @@ Object.keys(proxyTable).forEach(function (context) {
   if (typeof options === 'string') {
     options = { target: options }
   }
+  options.logLevel = 'debug'
```
Aladdin-ADD: can we add a comment, so users can have a choice to turn off?

Aladdin-ADD:
@MaxMilton  I'm fine with the the current behavior. maybe we can just add a comment, letting users have a choice to turn off?

MaxMilton:
That sounds like a decent way to do things. :+1:

thegatheringstorm:
Thanks!

PR #753, opened 2017-06-21
babel should keep code comments. (close #730)

### LINKED ISSUE #730: webpack dynamic import naming doesn't work
I used comment to define chunkname as follows:
```javascript
import(/* webpackChunkName: "chunk1" */'../components/chunk1.js')
```
then change webpack config  `output.chunkFilename: utils.assetsPath('js/[name].js')`

but the build result has no file named chunk1.js.

At first I thought it's a webpack bug. so I post an issue on webpack:
https://github.com/webpack/webpack/issues/4861

But then I made a simple project without vuejs-template , chunkname works fine.

So I think there must be something wrong in the config of this template, could you help me find out ?

My test project: 
https://github.com/MarvinXu/webpack-chunkname-test

many thx!

LinusBorg:
They are necessary for webpacks "magic comments" to work, e.g. with `import(/* webpackChunkName: "chunk1" */ './component.vue')`

### DIFF
--- a/template/.babelrc
+++ b/template/.babelrc
@@ -4,7 +4,6 @@
     "stage-2"
   ],
   "plugins": ["transform-runtime"],
-  "comments": false,
   "env": {
     "test": {
       "presets": ["env", "stage-2"],


LinusBorg:
Tests fail b/c of selenium not starting up. unit tests work, that's good enough in this case...

PR #755, opened 2017-06-22
Use webpack.prod.conf when NODE_ENV === production

batfink:
When setting NODE_ENV to production locally dev-server.js should use webpack.prod.conf.


MaxMilton:
Can you expand a bit on when this is useful? Why would we set `NODE_ENV = production` and run the dev server?

LinusBorg:
Sorry, I already merged another PR that did the same. Thanks for your contribution nontheless!

PR #756, opened 2017-06-22
pass cssSourceMap option to vue-loader

KagamiChan:
Currently `vue-loader` itself might produce sourcemap even if we set `cssSourceMap` option to `false` in config. 

After looking into the code I find that `cssSourceMap` option is not passed to `vue-loader` itself but only presents in `loaders`, and [this line](https://github.com/vuejs/vue-loader/blob/master/lib/style-compiler/index.js#L45) is truthy if `cssSourceMap` is `undefined`.

I think this is somewhat confusing and it requires extra options to make it completely disabled.

This PR therefore proposes a explicit `cssSourceMap` option for `vue-loader` passed in `vue-loader.conf.js`


LinusBorg:
Thanks!

PR #758, opened 2017-06-22
[WIP] Update: add htmlhint

### LINKED ISSUE #210: HTML linting

Aladdin-ADD:
fixes #210

### DIFF
--- a/template/build/webpack.base.conf.js
+++ b/template/build/webpack.base.conf.js
@@ -41,6 +41,12 @@ module.exports = {
       },
       {{/lint}}
       {
+        test: /\.html/,
+        enforce: 'pre',
+        loader: 'htmllint-loader',
+        exclude: /node_modules/
+      },
+      {
         test: /\.vue$/,
         loader: 'vue-loader',
         options: vueLoaderConfig
--- a/template/package.json
+++ b/template/package.json
@@ -38,6 +38,9 @@
     "eslint-friendly-formatter": "^3.0.0",
     "eslint-loader": "^1.7.1",
     "eslint-plugin-html": "^3.0.0",
+    "htmlhint-loader": "1.3.0",
+    "eslint-plugin-html": "^2.0.0",
+    "htmllint-loader": "^1.3.8",
     {{#if_eq lintConfig "standard"}}
     "eslint-config-standard": "^6.2.1",
     "eslint-plugin-promise": "^3.4.0",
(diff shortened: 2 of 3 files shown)


MaxMilton:
Should we provide a `.htmlhintrc` file with some preset defaults too?

MaxMilton:
Also, should this be an option together with ESLint or should it be a separate option during vue-cli init time? I'm not sure which is the better option to be honest. Thoughts anyone?

nickmessing:
@Aladdin-ADD, @MaxMilton, I think having a separate option would be better than combining it with eslint since the whole vue ecosystem is non-opinionated and highly configurable, even tho I think nobody will use just one of them.

template/build/webpack.base.conf.js
```diff
@@ -41,6 +41,12 @@ module.exports = {
       },
       {{/lint}}
       {
+        test: /\.html/,
+        enforce: 'pre',
+        loader: 'htmlhint-loader',
+        exclude: /node_modules/
+      },
```
nickmessing: I think this should be wrapped in `{{#lint}}...{{/lint}}` just like `eslint-loader` or have another key in `meta.js` something like `htmlLint`

template/package.json
```diff
@@ -37,6 +37,7 @@
     "eslint-friendly-formatter": "^2.0.7",
     "eslint-loader": "^1.7.1",
     "eslint-plugin-html": "^2.0.0",
+    "htmlhint-loader": "1.3.0",
```
nickmessing: Same as above comment

nickmessing:
I think, just like @MaxMilton said, there should be `.htmlhintrc` and it would be nice to add `htmllint src/**/*.vue` to `scripts` in `package.json` I think.

Aladdin-ADD:
I got an error running `npm run dev`, after adding this loader.
image

there are two issues here:
+ showing "Failed to compile with 1 errors", it is no much helpful to a user. can we change some config to show more info?
+ why this error? I just followed the htmlhint-loader's docs, did I forget something?

MaxMilton:
Looks like there might soon be a solution for the blank error in `friendly-errors-webpack-plugin`, see this PR: https://github.com/geowarin/friendly-errors-webpack-plugin/pull/52

Issue #748, opened 2017-06-18
is there a ci/test on the repo?


Aladdin-ADD:

LinusBorg:
this repository uses CircleCI, yes.

The projects created *from* this template don't have any CI solution preconfigured or anything, though.

Aladdin-ADD:
thanks for your respond so quickly:(

I got an error running test on my Windows 10.
![image](https://user-images.githubusercontent.com/13050025/27264793-4d0c779c-54b9-11e7-8c3c-ea7b0fc8adce.png)

Aladdin-ADD:
and seems contributors cannot see the test result for now. can we do something?

sudo-suhas:
> I got an error running test on Windows 10.

This might have something to do with git `core.autocrlf` setting. Read more [here](https://help.github.com/articles/dealing-with-line-endings/)

Aladdin-ADD:
fixed using `dos2unix test.sh`. thanks~

PR #761, opened 2017-06-23
Chore: add npm scripts: test

Aladdin-ADD:


nickmessing:
IMO it's better to add `#!/bin/sh` to `test.sh`, `chmod +x` it and then add `./test.sh` as a script because it's `sh` compatible and there is no need for `bash` per se.

nickmessing:
Just as option, maybe it's worth considering replacing `sh` with `shelljs` just like vue?

Aladdin-ADD:
@nickmessing up to this. I encountered some questions when running test on windows 10. https://github.com/vuejs-templates/webpack/issues/748#issuecomment-309307368

nickmessing:
@Aladdin-ADD, that is fixed by using `shelljs` which is cross-platform

LinusBorg:
@nickmessing is this good to merge for you?

nickmessing:
@LinusBorg, yes, I use only linux based operating systems but this solution will not work on windows. As long as nobody who works on this template uses windows it's fine but if there is requirement to support windows I'd highly recommend rewriting `test.sh` in `.js` with `shelljs`.

So it's your call, do you want to support windows or unix is okay? :wink:

Aladdin-ADD:
I think this is ready to merge. 🚀 
to replace sh to shelljs, we can finish it in another PR.

LinusBorg:
Leave have a look  at #1218. I took a bit more ambitious approach to improve the testing story for the template.

PR #773, opened 2017-06-28
add editor directories to .gitignore

eddyerburgh:
Add editor directories and files to .gitignore

- vs code directory
- idea directory
- visual studio files


eddyerburgh:
Oops, wrong .gitignore

PR #785, opened 2017-07-04
Fix Unknown plugin istanbul error when no unit test

huguangju:


LinusBorg:
Damn, merged into the wrong branch. reverted. need a fresh PR.

PR #788, opened 2017-07-09
Add npm script for fixing lint errors

sudo-suhas:
Adds an npm script to fix eslint errors/warnings which are auto-fixable.

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -11,7 +11,8 @@
     "unit": "cross-env BABEL_ENV=test karma start test/unit/karma.conf.js --single-run"{{/unit}}{{#e2e}},
     "e2e": "node test/e2e/runner.js"{{/e2e}}{{#if_or unit e2e}},
     "test": "{{#unit}}npm run unit{{/unit}}{{#unit}}{{#e2e}} && {{/e2e}}{{/unit}}{{#e2e}}npm run e2e{{/e2e}}"{{/if_or}}{{#lint}},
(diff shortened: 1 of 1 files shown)


sudo-suhas:
If this is approved, I can make a similar PR to vuejs-templates/pwa

MaxMilton:
I like this idea in general but how does it fix into the development workflow, when are devs meant to run this command? Is it something that devs are expected to script into their git commit hook? If so should we also add some documentation for this? Otherwise I suspect people won't use it.

nickmessing:
@MaxMilton, I use that as git commit hook since it fails if there are "unfixable" rules failing. I think it's worth a shot to add it as precommit hook but I don't know if we want to make it a part of `webpack` template. This should be bare minimal webpack template.

sudo-suhas:
TBH, I use the `lint` or `lint:fix` script commands quite rarely. I usually setup pre  commit hooks with `lint-staged`

sudo-suhas:
@LinusBorg Shall I close this?

LinusBorg:
@sudo-suhas I'm not sure about this. I add these to my projects as well, sometimes.

On the other hand, I don't want to add too many scripts to the package.json and blow up the list. it should be quite easy for anyone interested in this option to add it, and I'm not even sure if I want to have people use it that haven't even understood what it does...

sudo-suhas:
I understand @LinusBorg. Closing.

PR #792, opened 2017-07-10
Remove duplicate presets from .babelrc

guidobouman:
Babel environments automatically pick up the default settings. This way the coverage runs on the same output that a normal build would generate.


LinusBorg:
This was done intentionally. Because in unit tests we have to compile ES modules to commonjs.

PR #799, opened 2017-07-15
Enable cache for eslint-loader

sudo-suhas:
Docs - https://github.com/MoOx/eslint-loader#cache-default-false
> This option will enable caching of the linting results into a file. This is particularly useful in reducing linting time when doing a full build.

On a moderate size project, this brought down the production build time from ~60s to ~53s.

### DIFF
--- a/template/build/webpack.base.conf.js
+++ b/template/build/webpack.base.conf.js
@@ -36,7 +36,8 @@ module.exports = {
         enforce: 'pre',
         include: [resolve('src'), resolve('test')],
         options: {
-          formatter: require('eslint-friendly-formatter')
+          formatter: require('eslint-friendly-formatter'),
+          cache: true
         }
       },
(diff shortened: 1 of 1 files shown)


sudo-suhas:
ping

LinusBorg:
Note: I'm planning on updating all of the eslint dependencies soon, especially eslint 4. Could this have any influence on this feature? I don't think so...

sudo-suhas:
There's shouldn't be but, to be sure, I can test it once after eslint is updated.

MaxMilton:
I just struggled with issues caused by this change. Unless the cache is cleared after each change to the eslintrc file the changes are not respected to files already cached.

sudo-suhas:
My aplogies @MaxMilton. I'll try to make a PR to `eslint-loader` to fix this.

PR #801, opened 2017-07-16
Use const/let instead of var

sudo-suhas:
Use of `const`/`let` is considered [good practice](https://github.com/airbnb/javascript#references) and since the template specifies `"node": ">= 4.0.0"`, we can safely use these.

### DIFF
--- a/docs/backend.md
+++ b/docs/backend.md
@@ -6,7 +6,8 @@ Let's take a look at the default `config/index.js`:
 
 ``` js
 // config/index.js
-var path = require('path')
+'use strict'
+const path = require('path')
 
 module.exports = {
   build: {
--- a/docs/prerender.md
+++ b/docs/prerender.md
@@ -14,7 +14,7 @@ npm install --save-dev prerender-spa-plugin
 
 ```js
 // This line should go at the top of the file where other 'imports' live in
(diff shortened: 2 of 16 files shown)


sudo-suhas:
Rebased to fix conflict. Also added `'use strict'` directive because of this:
```
λ node --version
v4.0.0

λ echo let a = 1 > test-node4.js

λ node test-node4.js
E:\Projects\experiments\test-node4.js:1
(function (exports, require, module, __filename, __dirname) { let a = 1
                                                              ^^^
SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:413:25)
    at Object.Module._extensions..js (module.js:452:10)
    at Module.load (module.js:355:32)
    at Function.Module._load (module.js:310:12)
    at Function.Module.runMain (module.js:475:10)
    at startup (node.js:117:18)
    at node.js:951:3
```

sudo-suhas:
ping @LinusBorg. Any changes needed from my side?

LinusBorg:
Nope, was on vacation for 10 days, will do the merges this weekeend.

PR #809, opened 2017-07-18
Avoid no default ip error

bilougit:
https://github.com/karma-runner/karma-phantomjs-launcher/issues/84#issuecomment-168538895
If localhost is removed from /etc/hosts phantomjs won't start


LinusBorg:
host address will be customizable with #975 soon.

PR #820, opened 2017-07-25
Redirect everything to index.html

mikhailian:
It is not obvious that `connect-history-api-fallback` would not redirect a path with a dot in it. See a StackOverflow discussion here:
https://stackoverflow.com/questions/45280091/vue-router-does-not-catch-routes-with-a-dot-in-the-webpack-template/45296055

Let us make redirects slightly more explicit.

### DIFF
--- a/template/build/dev-server.js
+++ b/template/build/dev-server.js
@@ -52,7 +52,12 @@ Object.keys(proxyTable).forEach(function (context) {
 })
 
 // handle fallback for HTML5 history API
-app.use(require('connect-history-api-fallback')())
+app.use(require('connect-history-api-fallback')({
+  disableDotRule: true,
+  rewrites: [
+    {from: /\/app.js/, to: '/app.js'}
+  ]
+}))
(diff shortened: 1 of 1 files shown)


LinusBorg:
is this well tested with static files? Wouldn't this redirect requests for e.g. images to `index.html`?

mikhailian:
> Is this well tested with static files? Wouldn't this redirect requests for e.g. images to `index.html`?

Sure,  but anyone extending this template should configure web server routing, anyway. A rule that  looks at dot in the path to do it automagically just hides the complexity from developers.

I'd prefer to be hit with the need to configure the routing upfront rather than needing to figure out why some HTTP requests work and some don't.

I can suggest to use mime-types induced by Connect for images.  This is slightly better than the dot rule.

LinusBorg:
We will shortly be dropping the custom dev server in favor of `webpack-dev-server`, so this PR, is no longer applicable.

PR #831, opened 2017-08-02
delete root tag id `app`

g8up:
the id `app` is used in `index.html`


LinusBorg:
That element in index.html will be replaced with the one from App.vue when the app mounts. So this is intentional.

But nontheless, thanks for you willingness to improve this template!

PR #834, opened 2017-08-03
Fix standard eslint config link

afontcu:
Hi, just a minor tweak. The JavaScript Standard Style doc was moved so I updated the link accordingly.

Thanks for your work.


LinusBorg:
Thanks!

PR #837, opened 2017-08-04
Put hotMiddleware before proxyMiddleware in dev server

rhythnic:
Hot reload breaks when there are entries in the proxy table in config/index.js.  In the dev server, using the hot reload middleware before the proxy middleware fixes the problem.

### DIFF
--- a/template/build/dev-server.js
+++ b/template/build/dev-server.js
@@ -42,6 +42,10 @@ compiler.plugin('compilation', function (compilation) {
   })
 })
 
+// enable hot-reload and state-preserving
+// compilation error display
+app.use(hotMiddleware)
+
 // proxy api requests
 Object.keys(proxyTable).forEach(function (context) {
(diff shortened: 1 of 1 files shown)


LinusBorg:
Thanks for this. is there any info/docs that I can read up on this behaviour?

rhythnic:
Not that I know of.  I've never used `webpack-hot-middleware` or `webpack-proxy-middleware`.  When looking for documentation, I could only find [this tutorial](https://medium.com/@jamsesso/webpack-proxy-with-http-proxy-middleware-14331e7562d2).  He also puts the hot-middleware before the proxy middleware.  I've used `react-hot-loader`, and I know it is fickle about it's placement in the webpack setup, so I tried changing the order as a fix and it worked.

LinusBorg:
Thanks for the background info. Will try this!

Issue #751, opened 2017-06-20
webpack3 is coming


cccRaim:
so when upgrade？

MaxMilton:
I played with it a little bit and had problems with `webpack-hot-middleware` not doing hot updates correctly but rather reloading the whole page. Other than that things worked as expected, however, I didn't do any in-depth testing.

PanJiaChen:
@MaxMilton same problem
l remove 
```
// compiler.plugin('compilation', function (compilation) {
//     compilation.plugin('html-webpack-plugin-after-emit', function (data, cb) {
//         hotMiddleware.publish({action: 'reload'});
//         cb()
//     })
// });
```
now hot updates is ok

limichange:
How to fix this warning 🤔
```shell
⠧ building for production...(node:91029) DeprecationWarning: Chunk.modules is deprecated. Use Chunk.getNumberOfModules/mapModules/forEachModule/containsModule instead.
```

LinusBorg:
There's still some plugins that we rely on that have to release compatible versions (`` extract-text-webpack-plugin and `html-webpack-plugin` among them) so we will wait for those issue to resolve before we make the jump.

sudo-suhas:
@LinusBorg `extract-text-webpack-plugin` and `html-webpack-plugin` have been updated to support webpack 3. `inject-loader` has still not been updated though.

LinusBorg:
Do you know what issues exist with inject-loader with webpack 3?

sudo-suhas:
There was a [PR merged in](https://github.com/plasticine/inject-loader/commit/8be6a68e46c5382a45eb5d2c247866b5b925eb5b) which allowed webpack 3 peer dependency. But not released.

sudo-suhas:
@limichange I think this warning was coming from `extract-text-webpack-plugin`.

> DeprecationWarning: Chunk.modules is deprecated.

Looks to be fixed - [extract-text-webpack-plugin/CHANGELOG.md#breaking-changes](https://github.com/webpack-contrib/extract-text-webpack-plugin/blob/master/CHANGELOG.md#breaking-changes)

limichange:
@sudo-suhas thx!

MaxMilton:
I've opened an issue with inject-loader: https://github.com/plasticine/inject-loader/issues/46

I believe this is the last package now which needs updating before we can work on our own webpack 3 compatibility.

MaxMilton:
[...]

PR #844, opened 2017-08-07
Karma browserselection

c0defre4k:
Add an option to choose which browser should be used for Karma unit tests


c0defre4k:
I've removed the ModuleConcatentionPlugin from development config as its of course not needed there.
I also updated webpack to version 3. Yes, the title of the PR doesn't indicate that. Should I open another PR or just update the title of this PR?

template/package.json
```diff
@@ -34,7 +34,7 @@
     "css-loader": "^0.28.0",
     "cssnano": "^3.10.0",
     {{#lint}}
-    "eslint": "^3.19.0",
+    "eslint": "^4.4.0",
```
Aladdin-ADD: eslint just release a patch v4.4.1

c0defre4k: Thats right, but since the version string is **^**4.4.0 is should always install the newest available version < 5.0.0.

Aladdin-ADD: there is a small diffence that when has installed eslint@4.4.0.

MaxMilton:
@c0defre4k webpack 3 support is going to be a significant change and definitely should be handled in a separate PR. There's an issue with some background info: https://github.com/vuejs-templates/webpack/issues/751

c0defre4k:
@MaxMilton I see - thanks. I've updated the PR to focus on Karma browserselection

LinusBorg:
We will deprecate the karma setup in the forseeable future in favor of jest.

For that reason, I don't think it's  very useful to extend the questionaire with further options, so I have to close this PR.

Thanks for the efforts nontheless!

PR #854, opened 2017-08-13
Return exit code 1 when npm run build fails

pksunkara:
Having issues in CI's because of it returning 0 even when the build is invalid.


LinusBorg:
thanks!

PR #861, opened 2017-08-16
prioratize loading .vue over .js files

naderio:
this enables use of [vue-loader' Src Imports](https://vue-loader.vuejs.org/en/start/spec.html#src-imports) with the same file base name, e.g.:

`App.vue`
```html
<template> ... </template>
<style> ... </style>
<script src="./App.js"></script>
```
with code in `App.js`

### DIFF
--- a/template/build/webpack.base.conf.js
+++ b/template/build/webpack.base.conf.js
@@ -19,7 +19,7 @@ module.exports = {
       : config.dev.assetsPublicPath
   },
   resolve: {
-    extensions: ['.js', '.vue', '.json'],
+    extensions: ['.vue', '.js', '.json'],
     alias: {
       {{#if_eq build "standalone"}}
       'vue$': 'vue/dist/vue.esm.js',


Aladdin-ADD:
I don't think it is worth to change. actually if has `app.js` and `app.vue`, it make more sense to import `app.js`.

naderio:
@Aladdin-ADD prioritizing .js is certainly valid for entry point, still that is not the case for components (with src imports).

anyhow, a workaround for this issue would be to a different naming conversion, but that would be counter intuitive and not practical when used extensively (better have `Component.vue`, `Component.js`, `Component.css` than any other possible alternative), still the workaround would be ok when used just once (in your case just `App.vue`).

LinusBorg:
`.js` is the default file extension for any javascript app. I don't think that it's a good idea to change that, I feel it would be counter-intuitive for anyone who worked iwith webpack before.

This being a boilerplate there's nothing wrong with changing that if it better fits your workflow, but I think the default should be as close to a "normal" webpack config as possible.

naderio:
clear enough, closing

PR #863, opened 2017-08-17
Renamed build to build_config and change necessary references as 'build'…

vikashmadhow:
… is usually the output folder of some build tools and thus, in those cases, configured to be ignored by source control tools.


LinusBorg:
@chrisvfritz What do you think about this? Might make sense...

chrisvfritz:
I think it couldn't hurt, though I'd change build_config to build-config, since we're using kekab-case for all other folder and file names.

grakic:
How about this?

```
webpack/
├── build
│   ├── build.js
│   └── ...
├── index.js
├── dev.env.js
└── prod.env.js

```

MaxMilton:
`/build` is pretty standard across most JS projects. I'd see this as more of an edge case. May be best to manually modify the path after initialising your projects.

PR #870, opened 2017-08-22
add HashedModuleIdsPlugin when build

keepgoingwm:
use HashedModuleIdsPlugin to keep vendor.[hash].js same when third-party lib does not change
see more [webpack caching](https://webpack.js.org/guides/caching/#module-identifiers)


LinusBorg:
Thanks!

PR #871, opened 2017-08-22
Removing Vue.config.debug usage from documentation

adam-lynch:
It has been removed; https://vuejs.org/v2/guide/migration.html#Vue-config-debug-removed


LinusBorg:
Simply remoing it is not a good idea, we still want to have an example of how to use environment variables. Could you come up with a small one-liner like we currently do with the debug example?

adam-lynch:
@LinusBorg sure, updated

LinusBorg:
Perfect, thank you :)

PR #874, opened 2017-08-27
Switch from using Karma and PhantomJS to mocha-webpack and JSDOM

asselin:
This PR switches the test runner from Karma to mocha-webpack, and the DOM implementation from PhantomJS to JSDOM.  Coverage reports are generated using nyc.

Benefits:

- mocha-webpack has a watch mode to enable continuously running unit tests.  I added a new command to run them: `npm run unit:watch`
- JSDOM has a faster startup time and is a lighter weight alternative to PhantomJS, making unit tests run faster.


LinusBorg:
Hey, 

Thanks for this PR. However, we are planning to switch to JEst as the preferred test runner. We will keep the current option of running karma available for a while as people are used to it, but I doubt that we want to add a similar alternative like yours to the list.

However, we are in the planning stages for vue-cli 3 (see the vue-cli repository), and for that version, we plan on making something like different presets available. That might be a way to make this an optional "add-on" in a future version of our template system.

But I will close this PR because it's currently not a fit for where this template itself is going right now.

Issue #600, opened 2017-03-15
Autoprefixer is only active for .vue files


fvsch:
This seems to be the same bug as https://github.com/vuejs-templates/webpack/issues/85, which was closed because, I quote, "Autoprefixer is in the latest template release", but we don't get Autoprefixer working for CSS or SCSS files outside of `.vue` components.

Steps to reproduce:

```
vue init webpack postcss-test && cd postcss-test
npm install
npm install node-sass sass-loader --save-dev
```

Then create a few test stylesheets:

```css
/* src/assets/test1.css */
.test1 { display: flex; }
```

```scss
// src/assets/test2.scss
.test2 { display: flex; }
```

Finally, let's test a few ways to load these:

1. As entry points:

```js
// build/webpack.base.conf.js
...
module.exports = {
  entry: {
    app: './src/main.js',
    test1: './src/assets/test1.css',
    test2: './src/assets/test2.scss'
  },
  ...
}
```

Then `npm run build`. Results:

```css
/* dist/static/css/test1.ca0cd0e2b1b09d965f5ce5c21f1fc6e7.css */
.test1{display:flex}

/* dist/static/css/test2.4d716aeffa93f62596ef617b3129561c.css */
.test2{display:flex}
```

2. As JS imports:

```html
<!-- src/App.vue -->
<script>
import {} from './assets/test1.css'
import {} from './assets/test2.scss'
...
</script>
```

(Remove the entry points test before.) Then `npm run build`. Result:

```css
/* dist/static/css/app.4006533b8ed715dfd365cdd7a0b4f7be.css */
.test1,.test2{display:flex}
```

3. In style tags in components:

```html
<!-- src/App.vue -->
<style>
@import './assets/test1.css';
.test1b { display: flex; }
</style>

<style lang="scss">
@import './assets/test2.scss';
.test2b { display: flex; }
</style>
```

Result:

```css
/* dist/static/css/app.6c438b7dca17522bb707d18112b8cbcb.css */
.test1{display:flex}.test1b,.test2,.test2b{display:-webkit-box;display:-ms-flexbox;display:flex}
```

Conclusion: Autoprefixer is only used for style tags inside `.vue` components.
When using Sass or SCSS syntax and an external Sass/SCSS stylesheet, this can be used to apply Autoprefixer to code stored outside of components; the same trick does not work for CSS imports.

Was #85 closed based on a faulty assumption?

[...]

PR #890, opened 2017-08-31
fix postcss config

yibuyisheng:
fix [#600](https://github.com/vuejs-templates/webpack/issues/600)

### DIFF
--- a/template/build/utils.js
+++ b/template/build/utils.js
@@ -20,9 +20,13 @@ exports.cssLoaders = function (options) {
     }
   }
 
+  var postcssLoader = {
+    loader: 'postcss-loader'
+  }
+
   // generate loader string to be used with extract text plugin
   function generateLoaders (loader, loaderOptions) {
-    var loaders = [cssLoader]
+    var loaders = options.usePostCSS !== false ? [cssLoader, postcssLoader] : [cssLoader]
     if (loader) {
       loaders.push({
         loader: loader + '-loader',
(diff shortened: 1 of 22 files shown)


template/build/utils.js
```diff
@@ -20,9 +20,13 @@ exports.cssLoaders = function (options) {
...
   }
 
+  var postcssLoader = {
+    loader: 'postcss-loader'
+  }
+
   // generate loader string to be used with extract text plugin
   function generateLoaders (loader, loaderOptions) {
-    var loaders = [cssLoader]
+    var loaders = options.usePostCSS !== false ? [cssLoader, postcssLoader] : [cssLoader]
```
LinusBorg: 1. Where does `option.usePostCSS` come from? It's not to be found in `/config/index.js`, not in any of the webpack config files. This should be documented someway.

Is the idea to simply add it to the styleLoader code like this?

```javascript
// in webpack.base.conf.js:

rules: utils.styleLoaders({ sourceMap: config.dev.cssSourceMap, usePostCSS: true })
```

yibuyisheng: > Is the idea to simply add it to the styleLoader code like this?
>
> ```js
> // in webpack.base.conf.js:
>
> rules: utils.styleLoaders({ sourceMap: config.dev.cssSourceMap, usePostCSS: true })
> ```

Yes, I think the `usePostCSS` config should be set in the webpack config file.

PR #893, opened 2017-09-02
Remove unused fs require from webpack.base.conf.js

robwierzbowski:
I applied the linter script to config files (a good idea, IMO), and found this `fs` was unused.

Can someone with more familiarity with the project check that the cli doesn't insert anything into this file that uses `fs`? I'm pretty sure it doesn't.


LinusBorg:
It seems that this is a leftover from a recent change that I had to undo because it caused issue.

Thanks!

PR #894, opened 2017-09-02
Remove duplicate 'start' npm script

robwierzbowski:
The `start` and `dev` npm scripts are the same. Do you think it would be a good idea to standardize on a single task to run the dev server?

Originally added here: https://github.com/vuejs-templates/webpack/commit/6f7c03d85021c426aaa7465406a43bcf503a13d9

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -6,7 +6,7 @@
   "private": true,
   "scripts": {
     "dev": "node build/dev-server.js",
-    "start": "node build/dev-server.js",
+    "start": "npm run dev",
     "build": "node build/build.js"{{#unit}},
     "unit": "cross-env BABEL_ENV=test karma start test/unit/karma.conf.js --single-run"{{/unit}}{{#e2e}},
     "e2e": "node test/e2e/runner.js"{{/e2e}}{{#if_or unit e2e}},


MaxMilton:
I like this but developers new to the whole npm and node thing might expect to be able to run `npm start`. Perhaps start could just be `npm run dev`?

robwierzbowski:
Unless it's a convention in other webpack projects to have both `start` and `dev` I think it's a little redundant. The docs are pretty clear on how to start dev, build, etc. 

For broken expectations, I wonder if a user new to frontend but experienced in node would think `start` would run dev or build? Not sure.

If you think it's important to keep start, I like your idea of one task running the other so the actual dev task doesn't need to be copy pasted. I'll revise the PR.

robwierzbowski:
The more I think about it, the more I like the reference to `dev` in `start` :). Great suggestion.

template/package.json
```diff
@@ -6,9 +6,9 @@
   "private": true,
   "scripts": {
     "dev": "node build/dev-server.js",
-    "start": "node build/dev-server.js",
+    "start": "npm run dev",
     "build": "node build/build.js"{{#unit}},
-    "unit": "cross-env BABEL_ENV=test karma start test/unit/karma.conf.js --single-run"{{/unit}}{{#e2e}},
+    "unit": "cross-env BABEL_ENV=test karma dev test/unit/karma.conf.js --single-run"{{/unit}}{{#e2e}},
```
LinusBorg: I'm not aware of a `karma dev` command. What is that?

robwierzbowski: Nice code review. A mistake, fixing now :)

robwierzbowski:
Fixed. Thanks!

PR #897, opened 2017-09-03
Fix small errors in webpack.prod.conf

robwierzbowski:
I found a two small errors in webpack.prod.conf. One is an unused parameter, deleted in the commit below.

The second issue I need a little input on. [Here](https://github.com/vuejs-templates/webpack/blob/develop/template/build/webpack.prod.conf.js#L104) you're requiring a package that's not installed in the project by default. BUT the option that requires it is turned off by default. 

Do you think it's better to install the package, so users can turn the gzip option on and have everything work right away, or leave the package out of package.json, for quicker download? I think the first option is better, esp. for testing and maintenance of the project. 

Let me know what you think and I'll resolve this second issue and ping for code review.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,6 +1,6 @@
 # Introduction
 
-This boilerplate is targeted towards large, serious projects and assumes you are somewhat familiar with Webpack and `vue-loader`. Make sure to also read [`vue-loader`'s documentation](http://vuejs.github.io/vue-loader/index.html) for common workflow recipes.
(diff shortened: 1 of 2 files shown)


LinusBorg:
I think we will leave the gzip option as it is. `/config/index.js` has comments that tell users to install that package before setting that option to `true`, so I think this is covered well enough.

We usually like to keep the package.json as light as possible, it's already heavy enough.

PR #898, opened 2017-09-03
Simplify elementCount custom assertions

robwierzbowski:
This PR takes care of three things:

1. Slightly improves the file comment.
2. Renames one of two identically named variables that, because of separate scope, are actually pointing to different pieces of memory. The `selector` in the function passed to the browser is not the same data as the `selector` in brackets immediately after it. 
3. Calls a callback directly instead of using `.call()`.

Let me know what you think. #2 was especially confusing to me while I was looking at this file for the first time. If you can think of a better argument name, let me know 🙃.

### DIFF
--- a/template/test/e2e/custom-assertions/elementCount.js
+++ b/template/test/e2e/custom-assertions/elementCount.js
@@ -18,8 +18,8 @@ exports.assertion = function (selector, count) {
   }
   this.command = function (cb) {
     var self = this{{#if_eq lintConfig "airbnb"}};{{/if_eq}}
-    return this.api.execute(function (selector) {
-      return document.querySelectorAll(selector).length{{#if_eq lintConfig "airbnb"}};{{/if_eq}}
+    return this.api.execute(function (selectorToCount) {
+      return document.querySelectorAll(selectorToCount).length{{#if_eq lintConfig "airbnb"}};{{/if_eq}}
     }, [selector], function (res) {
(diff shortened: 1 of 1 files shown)


robwierzbowski:
Side note: I see a lot of templating to add semicolons. I'm converting all of my code to AirBnb style, and I think it might be possible to write all of the code in a way that it could be transformed to Standard or AirBnB with a `lint --fix` command. Have you tried that? It would be a lot cleaner than all of the `if AirBnB then insert ;` statements.

LinusBorg:
tbh I'm not super familiar with Nightwatch, but that change for the callback means that `this` is now undefined in the callback, which the original code probably intentionally avoided.

robwierzbowski:
From what I read in Nightwatch docs, the `this` binding isn't necessary in the callback. I looked through the history and there aren't any clues to why it was added. I've tested and everything seems to work as expected.

I'm most interested in making the arguments more clear (c3a2181), so if you'd like to keep the bound callback I can remove the commit that changes it.

LinusBorg:
@vuejs-templates/collaborators Can anybody with more nightwatch experience chime in wether it's imortant to bind `this` here?

yyx990803:
`this` will point to the browser instance, I believe it's a convention for most of the callbacks in nightwatch. I'd preserve it for the sake of compatibility.

LinusBorg:
@robwierzbowski There's your answer :)

robwierzbowski:
OK, commit removed. Very small PR now :)

PR #905, opened 2017-09-07
Remove unnecessary ternary in webpack prod config

robwierzbowski:
Another small potential cleanup. We don't need to check if we're in a testing environment and change the `env` string in `webpack.prod.conf`'s `webpack.DefinePlugin`. `webpack.test.conf` overrides the `prod` value of `webpack.DefinePlugin` when we're testing.

Or, is there a situation where we would want to run webpack with `webpack.prod.conf` settings, but have webpack think just the node env is `testing`?


LinusBorg:
Have a look at `/test/e2e/runner.js` - Here, we set NODE_ENV to 'testing', expecting to run the dev server with a production config.

robwierzbowski:
Ah I see. I assumed it would be run with the webpack.test.conf. Closing.

PR #911, opened 2017-09-07
added changing dev server's hostname via config/env var

ahmpro:
Changing dev server's hostname is useful for example while running inside Vagrant


LinusBorg:
This will be solved via #975. this PR won't be mergable after #975 because `dev-server.js` will be removed.

Sorry to close this :(

PR #913, opened 2017-09-08
Add dynamic port allocation

adi518:
Scans for an open port starting with the port set in `config`.

### DIFF
--- a/template/build/dev-server.js
+++ b/template/build/dev-server.js
@@ -84,6 +84,7 @@ devMiddleware.waitUntilValid(() => {
     if (err) {
       _reject(err)
     }
+    process.env.PORT = port
     var uri = 'http://localhost:' + port
     console.log('> Listening at ' + uri + '\n')
     // when env is testing, don't need open it
--- a/template/build/dev-server.js
+++ b/template/build/dev-server.js
@@ -72,14 +72,22 @@ var readyPromise = new Promise(resolve => {
   _resolve = resolve
 })
 
+var server
+var portfinder = require('portfinder')
+portfinder.basePort = port
+
 console.log('> Starting dev server...')
 devMiddleware.waitUntilValid(() => {
-  console.log('> Listening at ' + uri + '\n')
-  // when env is testing, don't need open it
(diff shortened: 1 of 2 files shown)


LinusBorg:
@chrisvfritz this would break e2e as nighwatch needs to know the port of the devserver to run the tests.

adi518:
Should I add a condition for testing environment?

LinusBorg:
That would be good, yes.

chrisvfritz:
@adi518 @LinusBorg Another potential option is we could set an environment variable here with the port number, that is later read by Nightwatch.

LinusBorg:
Great idea.

adi518:
I set the variable as suggested by @chrisvfritz.

chrisvfritz:
@LinusBorg How's this looking to you?

LinusBorg:
Will take a final look at the weekend when I'm back from portugal

chrisvfritz:
@LinusBorg Sorry, forgot you were on vacation! 😅

template/build/dev-server.js
```diff
@@ -68,22 +68,34 @@ app.use(staticPath, express.static('./static'))
...
-  _resolve()
+  portfinder.getPort((err, port) => {
+    if (err) {
+      _reject(err)
+    }
+    process.env.PORT = port
+    var uri = 'http://localhost:' + port
+    console.log('> Listening at ' + uri + '\n')
+    // when env is testing, don't need open it
+    if (autoOpenBrowser && process.env.NODE_ENV !== 'testing') {
```
LinusBorg: When we talked about adding a check for 'testing' environment, I didn't think about skipping the `opn(uri)` - I was thinking about skipping the whole portfinder logic.

The point is: for e2e tests, we currently try to access the dev-server through the port that is defined in the config. If the portfinder starts the def-server at a different port because that port is in use, e2e tests will connect to that wrong port anyway.

So either we find a way to tell the e2e test runner about the new found port, or we have to skip portfinder completely hen running e2e tests.

adi518: I thought we already did that. That is, updating the environment variable read by the e2e config.

PR #914, opened 2017-09-08
fix e2e nightwatch chrome option

erwin16:
avoid issues with Chrome browser during e2e tests


LinusBorg:
Hi @erwin16 !

What issues does this prevent?

LinusBorg:
I will close this since I don't know what the change implies and what problems it is meant to solve. 

We can re-open this PR anytime if additional information is provided.

PR #915, opened 2017-09-09
Make dev-server work with assetsPublicPath

grakic:
If assetsPublicPath is set in configuration options, dev-server should use it as path to index.html

### DIFF
--- a/template/build/dev-server.js
+++ b/template/build/dev-server.js
@@ -51,8 +51,18 @@ Object.keys(proxyTable).forEach(function (context) {
   app.use(proxyMiddleware(options.filter || context, options))
 })
 
+// redirect when assetsPublicPath is a sub-path
+if (config.dev.assetsPublicPath != '/') {
+  app.get('/', function(req, res) {
+    res.redirect(config.dev.assetsPublicPath)
+  })
+}
+
 // handle fallback for HTML5 history API
-app.use(require('connect-history-api-fallback')())
+app.use(require('connect-history-api-fallback')({
+  // use index.html in assetsPublicPath
+  index: path.posix.join(config.dev.assetsPublicPath, 'index.html')
+}))
 
 // serve webpack bundle output
 app.use(devMiddleware)


jdhartley:
This fixes a bug I was running into, where the dev server was returning

> Cannot GET /assetsPublicPath/

(where `assetsPublicPath` is the directory)

Thanks!

LinusBorg:
Hi!

Can you elaborate a bit about what situations and problems this prevents? I'm not entirely sure I see the problem this is fixing.

grakic:
Hi,

In the `config/index.js` there is this `assetsPublicPath` configuration option. By default it is set to a root path. I am working on the application that is deployed to a sub-directory, so in my project I set `assetsPublicPath` to /panel/. 

The middleware `connect-history-api-fallback` is responsible for routing requests to the main /index.html file. You can read more [here](https://github.com/bripkens/connect-history-api-fallback). Because of `assetsPublicPath` my main index.html file is not served as localhost:5000/index.html but instead it is served as localhost:5000/panel/index.html.

Without this change, when I access http://localhost:5000/panel/ the request is not rewritten to index.html and I get the error "Cannot GET /panel/". Configuring connect-history-api-fallback with the correct index path is fixing this problem.

For convenience this change also adds a  redirect from / to /panel (or whatever is set as `assetsPublicPath`). You can now run the server and access http://localhost:5000 which would then redirect to the app hosted at http://localhost:5000/panel/ which will show the index.html content.

Nothing is changed if `assetsPublicPath` is left at the default value. In that case no redirect is added and `connect-history-api-fallback` will be configured with /index.html which is the default setting.

LinusBorg:
@grakic We will be switching to `webpack-dev-server` (see #975).

I will leave this PR open, but set a "request changes" block. 

I would ask you to test this issue with webpack-dev-server (you can check the PR's branch out and test it locally, see #975, or after we officially merged it) and either close this PR if the issue doesn't persist, or adjust this PR to apply to the new setup.

grakic:
Hi @LinusBorg I can confirm that there are no issues with webpack-dev-server now. Only redirect is missing, I've pushed this as #980

PR #923, opened 2017-09-15
Make changing port from cli possible.

UncleBill:
Can pass optional argument to `npm run dev`. like:

`npm run dev -- --port 6060` run at port 6060.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,10 +18,13 @@ $ npm install -g vue-cli
 $ vue init webpack my-project
 $ cd my-project
 $ npm install
+# run at default port 8080
 $ npm run dev
+# run at port 6060
+$ npm run dev -- --port 6060
 ```
 
-If port 8080 is already in use on your machine you must change the port number in `/config/index.js`. Otherwise `npm run dev` will fail.
+You can change the port at `config/index.js` as well.
 
 ## What's Included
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -18,11 +18,12 @@ $ npm install -g vue-cli
 $ vue init webpack my-project
 $ cd my-project
 $ npm install
+# run at default port 8080
 $ npm run dev
+# run at port 6060
+$ npm run dev -- --port 6060
 ```
 
-If port 8080 is already in use on your machine you must change the port number in `/config/index.js`. Otherwise `npm run dev` will fail.
```
posva: As this is still true, can you leave something similar like:

```
You can change the port at `config/index.js` as well if you don't want to pass the option every time
```

UncleBill:
@posva  updated.

README.md
```diff
@@ -18,10 +18,13 @@ $ npm install -g vue-cli
...
+You can change the port at `config/index.js` as well.
```
posva: _if you don't want to pass the option every time_

posva:
Actually, this is not necessary once #913 is merged

UncleBill:
Well, port detection is more intelligent, but my patch is very light without extra dependency, it provides chance to specify a port anytime .

LinusBorg:
Hi @UncleBill 

Thanks for your PR. However, I just merged port detection, also you can use the PORT env variable as well as the config. That should provide more than enough options to configure this.

So I have to close this PR.

PR #925, opened 2017-09-17
Babel env preset

botmaster:
babel-preset-env: a preset that configures Babel for you.
[More here](http://2ality.com/2017/02/babel-preset-env.html)


botmaster:
Oups wrong repo

PR #951, opened 2017-10-06
switch hello-world tag to PascalCase

hootlex:


IamManchanda:
Sorry but I strongly disagree!

How was this merged without any talks!

IamManchanda:
Vue automatically converts pascal case `HelloWorld` to kebab-case `hello-world`
Then why??

template/src/App.vue
```diff
@@ -4,7 +4,7 @@
     {{#router}}
     <router-view/>
     {{else}}
-    <hello-world/>
+    <HelloWorld/>
```
IamManchanda: Instead of changing this why not tell people that Vue automatically converts 

HTML is preferred kebab-case by developers/designers worldwide!

LinusBorg:
Hi there!

There's no reason to panic, so calm down, take a deep brath and relax.

The change you are seeing is only *stylistic* in nature. We made this change *in this example component   because we recently released our Style guide , and in it, we recommend PasclaCase for Components templates of Single File Components(!). Read more about it [here](https://vuejs.org/v2/style-guide/#Component-name-casing-in-templates-strongly-recommended).

But this style is:

1. just a recommendation
2. not a technical change, nothing was broken or anything.

You can still do both, and always could:
```
<hello-world>this works!!</hello-world>
<HelloWorld>this works, too!!</HelloWorld>
``

**Comment by hootlex** [comment]
Thanks for detailed answer @LinusBorg. I also updated the description for future reference.

**Comment by IamManchanda** [comment]
I prefer this one because its HTML like!

```
<!-- Everywhere -->
<my-component></my-component>
```

and not

```
<!-- In single-file components and string templates -->
<MyComponent/>

<!-- In DOM templates -->
<my-component></my-component>
```

That said, The preference is totally personal
so yeah, I am happy and satisfied
 ... ohk no issue :D

Toilal:
It should at least be consistent, there's a `<router-view>` tag just above in the same file.

LinusBorg:
Thanks for the hint :(

PR #956, opened 2017-10-07
Fix use strict in main config

igtulm:
There is a small mistake in webpack main config file in use strict line. This pull request will fix it.


exarus:
Duplicate of #955

PR #957, opened 2017-10-07
Bump ESLint version 

c01nd01r:
Hi!
After last dependecy update, "eslint-config-airbnb-base" with "^12.0.1" requires "eslint": "^4.8.0".
At the current moment in package.json set "eslint": "^3.19.0". This breaks the building of the project.

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -33,7 +33,7 @@
     "copy-webpack-plugin": "^4.0.1",
     "css-loader": "^0.28.0",
     {{#lint}}
-    "eslint": "^3.19.0",
+    "eslint": "^4.8.0",
     "eslint-friendly-formatter": "^3.0.0",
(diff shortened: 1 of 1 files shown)


LinusBorg:
Hm, that's unfortunate. I wasn't ready to jump to eslint 4 since I wanted to test a bit if every plugin etc. is working well with it, especially eslint-plugin-vue, which will be released soon.

I will test that asap, and until then, will change the airbnb version back to `11.3`, which was till compatible with eslint `3`

c01nd01r:
I checked the work of eslint v 4.8.0 with the current set of "Standart" and "Airbnb" configs. It seems that everything works well. 
And yes, I probably should have created an issue. In any case, thanks for paying attention to this.

PR #965, opened 2017-10-10
Fix bug in dev-server when a proxyTable entry is a string

schettino72:
`options` can not be a `const` because it is modified if value is a string.

### DIFF
--- a/template/build/dev-server.js
+++ b/template/build/dev-server.js
@@ -51,7 +51,7 @@ app.use(hotMiddleware)
 
 // proxy api requests
 Object.keys(proxyTable).forEach(function (context) {
-  const options = proxyTable[context]
+  let options = proxyTable[context]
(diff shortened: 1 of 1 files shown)


LinusBorg:
Sorry, already fixed this on my own after seeing an issue about it :/

LinusBorg:
Also, it can't be a `let` either because this code is injected into the client and not transpiled, so we have to use ES5 (`var`)

schettino72:
`dev-server.js` file is injected into client? surprising...

schettino72:
I checked the code and saw no change regarding this. I saw a commit on `dev-client.js`... wondering if you got confused somehow.

LinusBorg:
Oh, right, I indeed got confused. sorry.

LinusBorg:
Thanks!

Issue #964, opened 2017-10-09
Try and replace custom dev-server with webpack-dev-server


LinusBorg:
Originally, back in the day, we came up with our own dev-server implementation because `webpack-dev-server` because at that time, it was lacking some features, like the error-overlay.

Since then, it hs improved significantly, and I'm not aware of any particular noteworthy advantage of your custom devserver over webpack's.

So I'll look into this a bit more, and if there really is no advantage left, I will work out a PR to switch the template to webpack-dev-server.

This will remove the maintainance burden for the dev server from this template and also makes it more approachable for people already familiar with `webpack-dev-server`.

### TODO
- [x] Check feature parity (sucess!)
- [x] change webpack config and package.json scripts to use webpack-dev-server 
- [x] Test this setup.
- [x] remove dead code.
- [x] review and close PRs / notify contributors of those PRs that may still target the now removed dev-server code.
- [ ] Review & test

HaiDongSong:
**#  I'm a rookie, and I follow a video to learn to do app, and I edit and modify the dev-server.js in the video. I don't have this document. What do I do?**

LinusBorg:
I don't know what you are told to change, so I can't say what that means with the new dev server.

PR #969, opened 2017-10-11
remove useless code

njleonzhang:


LinusBorg:
This won't make it as we will shortly switch to `webpack-dev-server` ( #964)

PR #971, opened 2017-10-11
show dev server url using server ip not localhost

njleonzhang:
someTimes we may need open the page in other device, it's better to show server url as `http://10.32.0.110:8080` than `http://localhost:8080`


LinusBorg:
Thanks, but we are in thge process of switching to `webpack-dev-server`, so this change won't make it.

There will be a config setting for this after the switch though.

Issue #369, opened 2016-11-17
opn(uri) cause error then launch dev server in environment without browser


VicarEscaped:
**Vue.js version**

2.0.7

**Steps to reproduce**

```
$ docker run -it --rm alpine:latest
$ apk add --no-cache nodejs
$ npm install vue-cli
$ vue init webpack test
$ cd test 
$ npm install
$ npm run dev
```

**What is Expected?**

App starts

**What is actually happening?**
dev server is not starting beacuse server try to [open window](https://github.com/vuejs-templates/webpack/blob/master/template/build/dev-server.js#L72)  with running app 

JS Error
(node:15) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Exited with code 3

LinusBorg:
Well, feel free to remove the line if that does not suit your usecase. It's a template, you should adjust it to your needs.

We feel that opening the window is a nice helper for most scenarios, so I don't see a reason to remove this.

peterwilli:
While I agree with @LinusBorg that we everyone is able to make the template fit to their needs, I think this is a bit more complicated than that. For instance, those of us with a Chromebook that run their dev environment in a chroot (with Crouton) will experience the same error as Chrome is not detected from within the chroot. While there are options to open Chrome anyway, I totally understand that Vue will not implement such platform-specific workarounds.

I do however think it's important that if there is no browser detected, the error will be ignored.

LinusBorg:
That behaviour is customizable since Nov 12, 2017 in `/config/index.js`

PR #946, opened 2017-10-03
Add config option for watch polling.

LinusBorg:
There seem to be [quite a lot of environments](https://github.com/vuejs-templates/webpack/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aclosed%20poll) (e.g. dropbox folders, docker containers, vagrant ...) in which `webpack-dev-middleware`'s default mode for file watching doesn't work correctly, and thus HMR fails.

In those scenarios, switching to a polling mode seems to fix most people's problems. 

This PR adds a new config option for the dev environment to switch to poll mode - this adds the following options to `webpack-dev-middleware`:
```javascript
watchOptions: {
  aggregateTimeout: 300,
  poll: true
},
```

PR #974, opened 2017-10-11
Handle unhandled promise rejecion when dev-server.js tries to open a …

gw0:
Fix for #369, that to silently ignores the promise rejection.

Without this `npm run dev` on nodejs v8.6.0 shows an annoying warning:

```
 DONE  Compiled successfully in 4890ms                                                                                      21:02:38

> Listening at http://localhost:8080

(node:1164) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 3): Error: Exited with code 3
(node:1164) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```


LinusBorg:
We will switch to `webpack-dev-server` very soon (#946), so this should not be an issue afterwards.

PR #979, opened 2017-10-12
Updated readme

alexanderjmold-kitty:


LinusBorg:
This PR updates much more than just the README and I don't think that was your intention.

PR #980, opened 2017-10-12
Redirect to index in webpack-dev-server if assetsPublicPath is set (close #915)

grakic:
When assetsPublicPath is set, index.html and other files are server from this path
and not from the root. Starting webpack-dev-serve in this case is showing a blank
page on the root url. This change adds a redirection from the root url to
assetsPublicPath for convenience.


template/build/webpack.dev.conf.js
```diff
@@ -29,6 +29,14 @@ const devWebpackConfig = merge(baseWebpackConfig, {
     quiet: true, // necessary for FriendlyErrorsPlugin
     watchOptions: {
       poll: config.dev.poll,
+    },
+    before: function(app) {
+        // Redirect to index.html in assetsPublicPath if path is not set to root
+        if (config.dev.assetsPublicPath && config.dev.assetsPublicPath != '/') {
```
LinusBorg: 1. I think `!==` is better, stylistically.
2. Just thinking: Could this break anything for a relative path ('./')?

LinusBorg: this has to be tested before merge.

LinusBorg (changes requested):

PR #987, opened 2017-10-16
Fix missing dependency bumps

LinusBorg:
...for:
* vue-loader
* vue-template-compiler


template/build/dev-client.js
```diff
@@ -1,7 +1,7 @@
 /* eslint-disable */
 'use strict'
 require('eventsource-polyfill')
-const hotClient = require('webpack-hot-middleware/client?noInfo=true&reload=true')
+var hotClient = require('webpack-hot-middleware/client?noInfo=true&reload=true')
```
Aladdin-ADD: why using `var`? it's in strict mode, using const is fine, and recommended!

LinusBorg: Because this piece of code is injected into the client bundle and not transpiled - which means it breaks in older browsers.

PR #993, opened 2017-10-18
Fix: unsupported flag --ignore-certificates-errors

anaganisk:
Issue in chrome browser, fixed with test-type flag


LinusBorg:
Hi, and thanks for the PR.

Could you explain in more detail what this is fixing, exactly? 

* What is happens without this fix?
* How does the fix work - is this flag set by default and this overwrites this default?

LinusBorg:
Closing this since I have no idea what it does or solves.

PR #997, opened 2017-10-20
replace `id` with `name` in chunk files, Fix #996

### LINKED ISSUE #996: /* webpackChunkName: "name" */ is working without the exact name.
My async route:
```javascript
const MainView = () => import(/* webpackChunkName: "group-foo" */ '@/components/MainView')
const OneView = () => import(/* webpackChunkName: "group-foo" */ '@/components/OneView')
const TwoView = () => import(/* webpackChunkName: "group-foo" */ '@/components/TwoView')
```

After I run build:

The bundle name is `0` rather than `group-foo`.

SidKwok:
Use `id` to name the chunk files will lost the `webpackChunkName` somehow. But replace it with `name` will solve this issue.

code:
```javascript
const MainView = () => import(/* webpackChunkName: "group-foo" */ '@/components/MainView')
const OneView = () => import(/* webpackChunkName: "group-foo" */ '@/components/OneView')
const TwoView = () => import(/* webpackChunkName: "group-foo" */ '@/components/TwoView')
```

use `id`:

use `name`:

without using any magic comments and it will fallback to `id`:

### DIFF
--- a/template/build/webpack.prod.conf.js
+++ b/template/build/webpack.prod.conf.js
@@ -25,7 +25,7 @@ const webpackConfig = merge(baseWebpackConfig, {
   output: {
     path: config.build.assetsRoot,
     filename: utils.assetsPath('js/[name].[chunkhash].js'),
-    chunkFilename: utils.assetsPath('js/[id].[chunkhash].js')
+    chunkFilename: utils.assetsPath('js/[name].[chunkhash].js')
   },
   plugins: [
     // http://vuejs.github.io/vue-loader/en/workflow/production.html


LinusBorg:
Thanks!

SidKwok:
Happy to help ! `:)`

c0defre4k:
As mentioned in my comment in #996 this leads to bigger JavaScript code.

> Note that these filenames need to be generated at runtime to send the requests for chunks. Because of this, placeholders like [name] and [chunkhash] need to add a mapping from chunk id to placeholder value to the output bundle with the webpack runtime. This increases the size and may invalidate the bundle when placeholder value for any chunk changes.

PR #1001, opened 2017-10-21
Renamed production command

Piterden:
To build dev we run `npm run dev` but to build prod we run `npm run build`?
Are you serious?

I am sure it MUST be:
```bash
$ npm run dev
$ npm run prod
```
But only if you want to increase the learning curve, you could leave it like now.

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -7,7 +7,7 @@
   "scripts": {
     "dev": "node build/dev-server.js",
     "start": "npm run dev",
-    "build": "node build/build.js"{{#unit}},
+    "prod": "node build/build.js"{{#unit}},
     "unit": "cross-env BABEL_ENV=test karma start test/unit/karma.conf.js --single-run"{{/unit}}{{#e2e}},
     "e2e": "node test/e2e/runner.js"{{/e2e}}{{#if_or unit e2e}},
     "test": "{{#unit}}npm run unit{{/unit}}{{#unit}}{{#e2e}} && {{/e2e}}{{/unit}}{{#e2e}}npm run e2e{{/e2e}}"{{/if_or}}{{#lint}},


LinusBorg:
> Are you serious?

I've seen countless projects that use `build` to actually "build" a the project, so yes, we are serious. :)

LinusBorg:
Sidenote: did a short search on Github:

* "npm run prod" : 7k results.
* "npm run build": 403k results.

Piterden:
What if I want to build dev version of a project?
Or dev version is not building, but constructing or scaffolding?

Piterden:
WordPress - About 1,710,000,000 results
Laravel - About 13,300,000 results

Does it mean I should learn WordPress, instead of Laravel? @LinusBorg

Piterden:
piterden 4500
linusborg 15900

You rock, man!!! :+1:

LinusBorg:
Can I ask why you are so upset about the name of a command?

Piterden:
Because the code is manipulation of names. What, besides the names, contains your code?

LinusBorg:
Okay, I'll bite since I have some time to kill.

> Because the code is manipulation of names. What, besides the names, contains your code?

I'm not entirely sure I understand what you mean - In case you mean "names in programming are important, it's what transports the meaning of the code we write", I completely agree. But these things are often subjective, and they can and should be discussed constructively.

Opening a PR with "Are you serious? I'm sure it MUST be: ...." basically suggest that we did something obviously stupid, which is not the *best* way to start a productive conversation with anyone.

My reply to this was meant to show you that using `npm run build` is *very, very* common in javascript projects, so it can't be something *obviously stupid*, but rather something very common, so we can assume that most people familiar with javascript projects already have a good intuition about what it does.

People who are *not* familiar with npm commands can read the templates [documentation](http://vuejs-templates.github.io/webpack/commands.html) to learn quick and clearly about what each command does - that provides a better learning curve than any command name could on it's own, I think.

> What if I want to build dev version of a project?
> Or dev version is not building, but constructing or scaffolding?

Well, of course we can have a constructive *discussio* about which name is probably more or less meaningful to someone at face value. But then we quickly end up in endless bikeshedding.

### Example

Your proposal was to change it to `npm run prod` - But since `npm run dev` runs a dev-server, would the user not assume that this also runs a server, but in prouction mode?

So we could rename *both* commands:
`npm run start-dev-server``
`npm run build-for-production`

That's more meaningfull, right? that's communicating *very* clearly what each command does.

But now people will complain about these commands being much too long to type them over and over. See where it goes?

### The Point

So, don't get me wrong: Command names can be discussed and optimized, for sure. Vbut that should be done in a civil manner and with a better argument than *"I am sure it MUST be:"*.

The current commands are commonly used commands in the npm ecosystem and are also well documented, so I don't see that Your initial PR had much of an argument.

Piterden:
Please excuse my rudeness.
Unfortunately, I haven't so much time to kill, but I have one idea about it:
```js
{
// ...
  "scripts": {
    // ...
    "prod": "npm run build",
    // ...
  },
// ...
}
```
It would make 410k for us...

Akryum:
Side note: [now.sh](https://zeit.co/) uses `npm run build` to build your deployments.

blake-newman:
But what if you have multiple environments such as staging?

You build for then environment, you don't prod for the environment. Also there is nothing stopping you from changing

Piterden:
You build for development, build for staging, or build for production, isn't it?

Frizi:
You don't build for development. You run the development server. There is no "building" action for dev environment, as opposed to all the rest of them.

Piterden:
I see `webpack(webpackConfig, cb)` inside `build.js` command (the only one action there)

Also, I see `const compiler = webpack(webpackConfig)` inside `dev-server.js` command, together with other commands.

Ah, dev-server can work with production webpack config if the environment would be production or testing.

Do you absolutely sure, I don't have build for development, @Frizi ?

May I ask you, what is `build`? I mean essentially...

Akryum:
The `build` command essentially produces compiled files in a `dist` folder, that you can then upload to a server.
The `dev` command doesn't produce any file and instead starts the webpack development server with development-specific features like watching files and hot-module-replacement.
This is all regardless of the value of `NODE_ENV`.

Akryum:
> I see webpack(webpackConfig, cb) inside build.js command (the only one action there)

This is not an action, it just creates an instance of webpack.

LinusBorg:
The dev server process builds the files in memory, so as far as the developer is concerned? The result of this process is only accessible though the dev server. 

The build process on the other hand actually produces a tangible output on the /dist folder. 

So the result of the two are very different, not just the environment flag differs. 

BTW we are now essentially bike's herding about the exact same things I mentioned in my "example" above... 😷

Frizi:
All the outcome in dev command happens to land inside webpack's memory-fs, and it's directly served from there via express server. Those files are unusable outside of this context, as those are very big. Usually those contain `eval`ed modules and inlined sourcemaps. You should never use that server as your production environment.

That's why we have a `build` step, a de-facto standard used by many deployment solutions as your default app building hook. It just produces a bunch of ready to serve files. That doesn't mean that it has to run in production though, as you might have many different environments distinct from your main prod servers. All of those targets will use the same script, but having a different value inside `NODE_ENV` environment variable.

In case this pattern doesn't fit well into your project (because nothing is a silver bullet), you are free to adjust the files produced by the template as you please.

Piterden:
That makes sense for me. Thank you for explaining.

What about environments with non-JS backends? Can vue-cli be used for such of?

Frizi:
That's what you have `build` for again. It just produces a set of files to serve, you don't need JS on server side to deal with them, it's only required on your build machine. I'm commonly using nginx to serve those files later, but you are free to use whatever solution you prefer. Github pages works just as well 😉

Piterden:
piterden 4500 -> 19000
linusborg 15900 -> 16000

However, me too)) :wink:

PR #1003, opened 2017-10-22
airbnb eslint config compatibility with vuex

### LINKED ISSUE #883: Eslint Airbnb config is not compatible with Vuex.
`no-param-reassign` rule makes the Airbnb Eslint config incompatible with Vuex. In Vuex parameters mutation is required - this is how Vuex works.

So In order to fix this issue, the following override is needed:

``` json
"no-param-reassign": [
  "error",
  {
    "props": true,
    "ignorePropertyModificationsFor": [
      "state",
      "acc",
      "e",
      "ctx",
      "req",
      "request",
      "res",
      "response",
      "$scope"
    ]
  }
],
```

---
PS
All properties except `state` are in the `ignorePropertyModificationsFor` array by default.

anubhav7495:
Added override to `no-param-reassign` rule to make airbnb eslint config compatible with vuex
Fixes #883

### DIFF
--- a/template/.eslintrc.js
+++ b/template/.eslintrc.js
@@ -44,6 +44,22 @@ module.exports = {
       'js': 'never',
       'vue': 'never'
     }],
+    // disallow reassignment of function parameters
+    // disallow parameter object manipulation except for specific exclusions
+    'no-param-reassign': ['error', {
+      props: true,
+      ignorePropertyModificationsFor: [
+        'state', // for vuex state
+        'acc', // for reduce accumulators
+        'e', // for e.returnvalue
+        'ctx', // for Koa routing
+        'req', // for Express requests
+        'request', // for Express requests
+        'res', // for Express responses
+        'response', // for Express responses
+        '$scope', // for Angular 1 scopes
+      ]
+    }],
     // allow optionalDependencies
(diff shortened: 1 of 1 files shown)


template/.eslintrc.js
```diff
@@ -44,6 +44,22 @@ module.exports = {
       'js': 'never',
       'vue': 'never'
     }],
+    // disallow reassignment of function parameters
+    // disallow parameter object manipulation except for specific exclusions
+    'no-param-reassign': ['error', {
```
LinusBorg: This would have to be wrapped in a handlebars conditional `{{#if_eq lintConfig 'airbnb'}}`

anubhav7495: It already is. Check line 41.

LinusBorg: Oh. Sometimes I'm stupid :)

LinusBorg: I'll merge this and then remove some of the node-spcific exceptions, because this is a frontend-template, and you are likely not gonna write express routes in it - that would be something for a SSR-focussed template.

LinusBorg:
(close/reopen to trigger CI)

PR #1004, opened 2017-10-22
Removed unnecessary options from .babelrc

anubhav7495:
Removed unnecessary `target.browsers` property from `babel-preset-env` as `browserslist` is already configured in `package.json`

### DIFF
--- a/template/.babelrc
+++ b/template/.babelrc
@@ -1,10 +1,7 @@
 {
   "presets": [
     ["env", {
-      "modules": false,
-      "targets": {
-        "browsers": ["> 1%", "last 2 versions", "not ie <= 8"]
-      }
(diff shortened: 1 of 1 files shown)


LinusBorg:
That makes sense, but I'm unsure in which place to keep it. It might fit better in `babelrc`, but I think I remeber that the `browserList` field in package.json can be used by other tools? Do you know anything about that?

anubhav7495:
yeah, `browserslist` field can be used by Autoprefixer, Stylelint & babel-preset-env.

LinusBorg:
Thanks. Will merge this tonight when I do some cleanup.

LinusBorg:
Thanks!

PR #1007, opened 2017-10-24
Transform environment to ES6

VanMess:
Hey guys.

I had transform build and config code into ES6, I think it would be better for user to customized.


LinusBorg:
Hi, and thanks for this PR.

Unfortunately I don't think we want to focus on that right now. We also attempt to keep this template backward compatible to at least Node 6, which still didn't fully support ES6

babel-register allows to get around that, but it slows the process down, so it's not really an improvement from my POV.

PR #1008, opened 2017-10-25
Add standard package for standard lint option

dennisreimann:
This allows for nicer integration with editor plugins (i.e. [VS Code Standard Style](https://marketplace.visualstudio.com/items?itemName=chenxsan.vscode-standardjs).

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -43,6 +43,7 @@
     "eslint-plugin-standard": "^3.0.1",
     "eslint-plugin-import": "^2.7.0",
     "eslint-plugin-node": "^5.2.0",
+    "standard": "^10.0.3",
     {{/if_eq}}
     {{#if_eq lintConfig "airbnb"}}
     "eslint-config-airbnb-base": "^11.3.0",


LinusBorg:
What exactly does this do? "nicer integration" doesn't say much. I started writing changelogs so it might be helpful to mention what this actually helps with.

dennisreimann:
If you don't have standard installed globally, the VS Code Standard plugin complains about not being able to find the package (if it's not installed as a `devDependency`). 

A question I had in mind before proposing this: Does it make sense to do linting via ESLint in case one uses the `standard` option? To me it feels a bit redundant.

LinusBorg:
Thanks, that's clearer.

Well, this template doesn't assume people are using a spoecific editor or editor-plugin. IF people do, it's their responsibility to decide what do use and how to make it work.

So I would say that I don't think adding the `standard` package makes too much sense, since using both this and eslint would be uncesessary, would you agree?

dennisreimann:
Yeah, that's alright :)

Issue #1005, opened 2017-10-23
Object destructuring in v-bind


maurop123:
Passing props like this...
```
<puppy v-bind="{ ...dogLikeQualities }"></puppy>
```
Produces this error...
```
Module build failed: Error
    at Node.transpile (/home/mauro/code/sandbox/webpack/node_modules/vue-template-es2015-compiler/buble.js:12676:11)
    at /home/mauro/code/sandbox/webpack/node_modules/vue-template-es2015-compiler/buble.js:5864:59
    at Array.forEach (<anonymous>)
    at Node.transpile (/home/mauro/code/sandbox/webpack/node_modules/vue-template-es2015-compiler/buble.js:5864:10)
    at Node.transpile (/home/mauro/code/sandbox/webpack/node_modules/vue-template-es2015-compiler/buble.js:6553:28)
    at /home/mauro/code/sandbox/webpack/node_modules/vue-template-es2015-compiler/buble.js:5864:59
    at Array.forEach (<anonymous>)
    at Node.transpile (/home/mauro/code/sandbox/webpack/node_modules/vue-template-es2015-compiler/buble.js:5864:10)
    at Node.transpile (/home/mauro/code/sandbox/webpack/node_modules/vue-template-es2015-compiler/buble.js:6553:28)
    at /home/mauro/code/sandbox/webpack/node_modules/vue-template-es2015-compiler/buble.js:5864:59
    at Array.forEach (<anonymous>)
    at Node.transpile (/home/mauro/code/sandbox/webpack/node_modules/vue-template-es2015-compiler/buble.js:5864:10)
    at Node.transpile (/home/mauro/code/sandbox/webpack/node_modules/vue-template-es2015-compiler/buble.js:5991:28)
    at /home/mauro/code/sandbox/webpack/node_modules/vue-template-es2015-compiler/buble.js:5864:59
    at Array.forEach (<anonymous>)
```
Here's a project freshly built from this template that demos this issue: https://github.com/maurop123/object-destructure

The master branch only has the code changes replicating this issue, and the dev branch has the diff attempts to resolve it.

LinusBorg:
Vue-loader uses buble under the hood to get rid of the `with this { ... }` code that the template compiler produces.

You probably have to set the `objectAssign` option for buble in the vue-loader config as shown here:

https://vue-loader.vuejs.org/en/options.html#buble

Note that, as mentioned in that link, you still have to provide a polyfill for Object.assign to support older browsers (IE)

[...]

PR #1017, opened 2017-10-31
Support for Object destructuring in v-bind

maurop123:
Adds buble config to vue-loader.conf.js

Fixes this: https://github.com/vuejs-templates/webpack/issues/1005

### DIFF
--- a/template/build/vue-loader.conf.js
+++ b/template/build/vue-loader.conf.js
@@ -15,5 +15,11 @@ module.exports = {
     source: 'src',
     img: 'src',
     image: 'xlink:href'
+  },
+  buble: {
+    objectAssign: 'Object.assign',
+    transforms: {  // turn off the `with` removal
+      stripWith: false
+    }
   }
 }


LinusBorg:
I'd like to merge this, but for that we would have to include a polyfill for `Object.assign()` as well.

I'm not against that, but I have a general re-evaluation on the map about of what kind of polyfills we include, so this will be merged sometime, but might take a small while.

Thanks for your patience.

template/build/vue-loader.conf.js
```diff
@@ -15,5 +15,11 @@ module.exports = {
     source: 'src',
     img: 'src',
     image: 'xlink:href'
+  },
+  buble: {
+    objectAssign: 'Object.assign',
+    transforms: {  // turn off the `with` removal
+      stripWith: false
```
LinusBorg: That has to go definitely, it's not required to make object rest spread work - to the contrary, it would probably breka most builds using ES modules in webpack 2 and upwards.

maurop123:
Understood. In case it helps your evaluation of polyfills, we recently merged a PR on the Vuetify PWA template that added the `babel-polyfill` npm package to handle an issue with the service worker and vue router when for both hash bang mode and history mode.

Here's that PR for reference: https://github.com/vuetifyjs/pwa/pull/3/files

template/build/vue-loader.conf.js
maurop123: I don't understand. Are you saying that the buble block here should NOT be added because it would break most builds? I thought this was what you were suggesting [here](https://github.com/vuejs-templates/webpack/issues/1005#issuecomment-340443636) to make that issue work.

maurop123: OK I just read your comment [here](https://github.com/vuejs-templates/webpack/issues/1005#issuecomment-343876559). Is this why you're saying that adding this will break most builds?

It seems to suggest that projects built with recent version of the vuejs webpack template should not break. But it is breaking with the current version of the template, at least it was at the time I made [that issue](https://github.com/vuejs-templates/webpack/issues/1005).

LinusBorg: What I mean is that this is the option we want:
```
objectAssign: 'Object.assign',
```
and this is the option we don't want:
```
transforms: {  // turn off the `with` removal
+      stripWith: false
```

LinusBorg:
Thanks, will have a look!

LinusBorg:
Outdated

PR #1026, opened 2017-11-02
Use webpack 3's standard way to generate a manifest in CommonsChunkPlugin

mgenware:
The doc URL: https://webpack.js.org/plugins/commons-chunk-plugin/#manifest-file

* `minChunks` is set to `Infinity` so that only webpack runtime is written to manifest file.
* `chunnks: ['vendor']` is removed as entry chunks are selected if `chunks` is omitted.

These properties are especially useful when using multiple `CommonsChunkPlugin`s.

### DIFF
--- a/docs/e2e.md
+++ b/docs/e2e.md
@@ -10,7 +10,7 @@ Let's take a look at the files in the `test/e2e` directory:
 
 - `nightwatch.conf.js`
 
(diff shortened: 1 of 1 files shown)


mgenware:
@LinusBorg Just noticed you added a label, sorry for the duplicate PR, I accidentally deleted the remote branch, it was automatically closed now, have to start a new one #1028

PR #1027, opened 2017-11-02
Set ExtractTextPlugin.allChunks to true

mgenware:
When using CommonsChunkPlugin and there are extracted chunks (from ExtractTextPlugin.extract) in the commons chunk, allChunks must be set to true。

Doc: https://github.com/webpack-contrib/extract-text-webpack-plugin

### DIFF
--- a/template/build/webpack.prod.conf.js
+++ b/template/build/webpack.prod.conf.js
@@ -42,7 +42,8 @@ const webpackConfig = merge(baseWebpackConfig, {
     }),
     // extract css into its own file
     new ExtractTextPlugin({
-      filename: utils.assetsPath('css/[name].[contenthash].css')
+      filename: utils.assetsPath('css/[name].[contenthash].css'),
+      allChunks: true,
     }),
     // Compress extracted CSS. We are using this plugin so that possible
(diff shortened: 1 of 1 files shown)


mgenware:
For a quick testing of multiple common chunks, [this project](https://github.com/mgenware/fx63) may be helpful.

LinusBorg:
Hey, so I looked into this, and I don't think that I want to activate this by default. 

### Reasoning

With the current config, CSS from .vue files that are code-splitted will reside in that async chunk and only be loaded when that async chunk is actually loaded.

That means that the `app.css` file will be lighter and only contain the "basic" styles, and stuff you load later comes with its own styles.

### Solution

I think this option can be useful though, and so I would suggest to add it to the templates config like this:
```javscript
// set the following options to `true` if you want to extract CSS from
// codesplit chunks into this main css file as well.
// this will result in *all* of your app's CSS being loaded upfront.
allChunks: false
```

If you want to make this change to your branch, I will merge this so you get credit for bringing this to our attention.s

mgenware:
Hi @LinusBorg, thanks for your informative reply. I totally agree, I've committed changes to my branch 😀

PR #1028, opened 2017-11-02
Use webpack 3's standard way to generate a manifest in CommonsChunkPlugin

mgenware:
Use webpack 3's standard way to generate a manifest in CommonsChunkPlugin.

The doc URL: https://webpack.js.org/plugins/commons-chunk-plugin/#manifest-file
* `minChunks` is set to `Infinity` so that only webpack runtime is written to manifest file.
* `chunnks: ['vendor']` is removed as entry chunks are selected if `chunks` is omitted.

This is very useful when using multiple `CommonsChunkPlugin` plugins.


mgenware:
For a quick testing of multiple common chunks, [this project](https://github.com/mgenware/fx63) may be helpful.

Issue #546, opened 2017-02-23
Set the baseUrl for the dev-server


grantcarthew:
I just spent many hours trying to work out how to mount the dev server at a child path in the URL.

For example; my application is mounted at a child path of `/app/`. The root of the url `http://someaddress.com/` is owned by other code and the Vuejs app needed to be mounted under `http://someaddress.com/app/`

Finally I found the solution which is two part:

1.   Change the `assetsPublicPath` in the `config/index.js` file like so:

```js

  dev: {
    env: require('./dev.env'),
    port: 2001,
    autoOpenBrowser: true,
    assetsSubDirectory: 'static',
    assetsPublicPath: '/app/',
    proxyTable: {},
    // CSS Sourcemaps off by default because relative paths are "buggy"
    // with this option, according to the CSS-Loader README
    // (https://github.com/webpack/css-loader#sourcemaps)
    // In our experience, they generally work as expected,
    // just be aware of this issue when enabling this option.
    cssSourceMap: false
  }

```

2.   In the `dev-server.js` file, change the `connect-history-api-fallback` middleware to support a new index path:

```js

app.use(require('connect-history-api-fallback')({
  index: '/app/index.html'
}))


```

I suggest a change to the template to make it easier to change the `base` url used for mounting the app. I quick example of the change being something like this:

```js

app.use(require('connect-history-api-fallback')({
  index: config.dev.publicPath + '/index.html'
}))

```

That dynamic path would need to be parsed but you get the idea. Maybe behind a flag of `changeBaseUrl` or something like that.

grantcarthew:
I've been thinking on this last night. Rather than some multi-option like `changeBaseUrl` it would be better just to expose the index file option. Something like this:

```js

{
  defaultDocument: `/index.html`
}

```

This then allows the user to set the value to anything rather than just change the sub-path. Couple this with a good documentation example of how to change the Vue application to a sup-path and you are set.

LinusBorg:
Note: index.html is also referenced here:

https://github.com/vuejs-templates/webpack/blob/master/template/build/webpack.dev.conf.js#L29

and here:

https://github.com/vuejs-templates/webpack/blob/master/template/build/webpack.prod.conf.js#L54

grantcarthew:
Ah, thanks @LinusBorg.
Documentation on those to extra references: https://github.com/jantimon/html-webpack-plugin

I don't think I need to change those setting in my use case however I'm not sure how you would want to expose them to a template user through configuration.

[...]

PR #975, opened 2017-10-11
Switch to webpack-dev-server (close #964)

LinusBorg:
### Why?

The reasons for this change are laid out in #964 

### What changed?

- `/config/dev-server.js` and `/config/dev-client.js` are gone
- `webpack.dev.conf.js` now returns a promise. This was necessary to keep the recently added portfinder feature working without writing a new `dev-server.js` file that wraps `webpack-dev-server`
- That file now also has all the `webpack-dev-server` configuration, but most of the usually used options can be configured from `/config/index.js`
- `/test/e2e/runner.js` was changed quite a bit to make it work with `webpack-dev-server`, but it works just fine.

### What was added?

This PR also has a few goodies
- The host&port are now shown in the console after each build with a nice message
bildschirmfoto 2017-10-12 um 00 02 25

- build progress is shown in the console during each build in development (already worked in `npm run build`).
-  You can now optionally receive a notification from your OS when the build fails, this can be activated with the `notifyOnErrors` option in `/config/index.js`:
bildschirmfoto 2017-10-12 um 00 03 43
- You can deactivate eslint-loader from `/config/index.js`. This is useful if you already have eslint support in your editor/IDE and don't want the warnings /errors shown by webpack
- If you do use eslint-loader, it usually shows its errors in the error overlay in the browser and halts the build. You can now opt out of this, so eslint errors & warnings are only shown in the console.
- General restructuring of `/config/index.js`:
  - moving all dev options to the top since these are the ones you use first/more often
  - added more comments and links to relevant webpack docs where I deemed it useful.
  - and tried to group the options into more logical blocks

### Testing this PR

@vuejs-templates/collaborators and anyone interested: Please take this for a test-ride with:
```
vue init webpack#feature/webpack-dev-server(#964)
```

And report any and all problems in this PR thread. Thanks!

PR #1032, opened 2017-11-07
fix for (#546) Make public path the app's context

scoobah36:
setting the config:
```
assetsPublicPath: '/appContext/',
```
results in the app deploying all of it's content under that context 
```
http://localhost:8080/appContext/#/
```
all hot reloading and static asses will be served under that context

### DIFF
--- a/template/build/dev-client.js
+++ b/template/build/dev-client.js
@@ -1,7 +1,7 @@
 /* eslint-disable */
 'use strict'
 require('eventsource-polyfill')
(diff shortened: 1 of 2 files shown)


LinusBorg:
Thanks for your PR, but we are in the process of dumping the whole devserver setup (#975) - will merge this tonight and release on the WE I think.

so this won't be of any use, unfortunately.

monsdroid:
for me it is atm. thanks scoobah36!

PR #1039, opened 2017-11-10
Add sourcemap option to postCSS-loader options

LinusBorg:


LinusBorg:
Hm, thanks for testing it out - https://github.com/vuejs/vue-loader/pull/937 might also be of interest.

Seems some digging will be necessary to make this work. A job for the weekend.

LinusBorg:
@Tirke I found the issue - it was another mistake in the original PR, a strict `false` comparison even though `options.usePostCSS` is `undefined`when creating the loaders for vue-loader. 

That resulted in postss-loader being included where it shouldn't.

The fix I just pushed should solve this.

LinusBorg:
Good eye, but that's cosmetics. Will add this later.

PR #1044, opened 2017-11-12
Add historyApiFallback in dev-server config

Tirke:
I think this is an important one for a Vue SPA. Also it was used in the previous version of the dev server.

There is also a common addition to historyApiFallback which is :
```
historyApiFallback: {
      disableDotRule: true
}
```

But I don't think that setting this one by default is right.

### DIFF
--- a/template/build/webpack.dev.conf.js
+++ b/template/build/webpack.dev.conf.js
@@ -17,6 +17,7 @@ const devWebpackConfig = merge(baseWebpackConfig, {
   
   // these devServer options should be customized in /config/index.js
   devServer: {
+    historyApiFallback: true,
     hot: true,
     host: process.env.HOST || config.dev.host,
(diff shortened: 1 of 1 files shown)


LinusBorg:
Hi and thanks for the PR.

I remember discussions abotu this rule - however, we do not have this enabled in the old dev-server (which is still public in the `master` branch, and I agree that activating it by default now is not a good idea.

So we should think about if we want to add this to the `/config` options, or directly into the devserver-config, but commented accordingly and commented out so it's inactive.

Tirke:
We did have [that piece of config](https://github.com/vuejs-templates/webpack/blob/504bcdabe287bfb337483f964a6632d3953c657d/template/build/dev-server.js#L62) which is equivalent to just setting `historyApiFallback: true` no ?

PR #1053, opened 2017-11-13
Make sure that imported css files in SFCs are autoprefixed, too

LinusBorg:
When using postcss-loader, we have to adjust the `importLoaders` setting in css-loader to re-apply the previous loaders on imported files.

Without this setting, imports in css files will be read as plain css instead of being re-run through postcss, which I think is undesirable.

This is my first, untested attempt.

If any postcss(-loader) guru can chime in, help is welcome.

### DIFF
--- a/template/build/utils.js
+++ b/template/build/utils.js
@@ -14,11 +14,14 @@ exports.assetsPath = function (_path) {
 exports.cssLoaders = function (options) {
   options = options || {}
 
-  const cssLoader = {
-    loader: 'css-loader',
-    options: {
-      minimize: process.env.NODE_ENV === 'production',
-      sourceMap: options.sourceMap
+  function createCssLoader({ importLoaders }) {
+    return {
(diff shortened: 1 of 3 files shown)


template/build/utils.js
```diff
@@ -31,6 +34,10 @@ exports.cssLoaders = function (options) {
 
   // generate loader string to be used with extract text plugin
   function generateLoaders (loader, loaderOptions) {
+    let importLoaders = 0
+    if (options.usePostCSS) importLoaders++
+    if (options.usePostCSS && loader) importLoaders++
```
LinusBorg: actually I think we don't need this since preprocessors (SASS, LESS, Stylus) will handle `@import` themselves, so there won't be any left for css-loader to care about.
```javascript
const importLoaders = options.usePostCSS && !loader ? 1 : 0
```
that should be enough.

PR #1055, opened 2017-11-13
Remove logo.png inside build folder

mul14:
The `build/logo.png` is not necessary.


LinusBorg:
it is, for the notification feature:

https://github.com/vuejs-templates/webpack/blob/develop/template/build/utils.js#L96

PR #1061, opened 2017-11-14
fix webpack-dev-server can't access localhost

xsdlr:
I was unable to access the dev server on localhost:8080 on Mac OSX.  i see some suggest in webpack/webpack-dev-server#183.


LinusBorg:
That usually works, it has been the default setting forever, and worked in all of our tests.

There must be something specific to your situation, so I don't see a need to change the default.

Issue #1070, opened 2017-11-16
The console error log does not print correctly when there is an unknown variable in the index.html template


wolfstark:
Because I accidentally annotated a code, the error log can not be printed correctly, I can not find the source of the BUG.

```js
    new webpack.DefinePlugin({
      'process.env': require('../config/dev.env')
      // doman: JSON.stringify('http://cm.cn'),
    }), 
```
![image](https://user-images.githubusercontent.com/15010463/32876209-bc852c70-cad6-11e7-9fcd-801edfb02518.png)
The print result I expect is below, so that I can find the problem.

 error：Template execution failed: ReferenceError: doman is not defined

LinusBorg:
I don't understand. What does index.hyml have to do with it, or the comment in that code?

LinusBorg:
Okay, saw the PR and understand the issue now. Thanks!

LinusBorg:
merged

PR #1071, opened 2017-11-16
Fix # 1070 When index.html has a positional variable, the error can b…

wolfstark:
Now the error can be printed correctly.
image

Fixes this: [#1070](https://github.com/vuejs-templates/webpack/issues/1070)

### DIFF
--- a/template/build/utils.js
+++ b/template/build/utils.js
@@ -87,7 +87,7 @@ exports.createNotifierCallback = function () {
     }
     const error = errors[0]
 
-    const filename = error.file.split('!').pop()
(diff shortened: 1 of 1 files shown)


template/build/utils.js
```diff
@@ -87,7 +87,7 @@ exports.createNotifierCallback = function () {
     }
     const error = errors[0]
 
-    const filename = error.file.split('!').pop()
+    const filename = error.file && error.file.split('!').pop()
```
LinusBorg: good catch

PR #1072, opened 2017-11-16
[enhancement]Remove useless files and performance enhancement in dev server.

SidKwok:
1. Fix comments in `config/index.js`
`// can be overwritten by process.env.HOST` -> `// can be overwritten by process.env.PORT`

2.  remove env in `.babelrc` when not using unit or e2e

3.   remove `webpack.test.conf.js` when not using unit or e2e

4.   remove extra space in `.babelrc`
before:
```json
"env": {
    "test": {
      "presets": ["env", "stage-2"]    }
  }
```

after:
```json
"env": {
    "test": {
      "presets": ["env", "stage-2"]
    }
  }
```

5.  enable `compress` in devServer by default, it will be useful when you try to access your website in a remote server.

before:
3aa68ba82a2e113d693131789665ab7c

after:
211c9ce5d45785991a023ce010f96a47


meta.js
```diff
@@ -105,6 +105,7 @@ module.exports = {
     ".eslintrc.js": "lint",
     ".eslintignore": "lint",
     "config/test.env.js": "unit || e2e",
+    "build/webpack.test.conf.js": "unit || e2e",
```
LinusBorg: I think that config is only required if `e2e || (unit && runner === 'karma')` - jest doesn't use webpack.

SidKwok:
Done. @LinusBorg

LinusBorg:
Thanks. I will test this locally tonight and then approve & merge.

Issue #1062, opened 2017-11-14
Jest not able to find test specs in Windows 10


nirkki:
Fresh project with 'vue init webpack just-testing' doesn't find the test files in Win 10
...
  25 files checked.
  testMatch: **/__tests__/**/*.js?(x),**/?(*.)(spec|test).js?(x) - 1 match
  testPathIgnorePatterns: \\node_modules\\ - 25 matches
Pattern: test/unit/specs - 0 matches

LinusBorg:
@eddyerburgh Got an idea?

eddyerburgh:
@nirkki can you try running:

```
npm run unit -- --no-watchman
```

I'm basing this off a [stackoverflow answer](https://stackoverflow.com/questions/45132566/how-to-fix-jest-no-tests-found-on-windows-10).

fedegar33:
Same issue on Windows 7.  Tried to run `npm run unit -- --no-watchman` but the problem persists.

nirkki:
Didn't help here either. I also found some notes that Jest has had some issues with Windows paths containing special chars - but my project doesn't have those.

nirkki:
Tried to troubleshoot this further: installing jest globally didn't make difference on 'npm run unit'
But it enabled me to run it from the command line - and here is the result
```
C:\test\jesting> jest --coverage
 PASS  test\unit\specs\HelloWorld.spec.js
  HelloWorld.vue
    √ should render correct contents (27ms)

Test Suites: 1 passed, 1 total
Tests:       1 passed, 1 total
Snapshots:   0 total
Time:        1.403s
Ran all test suites.
-----------------|----------|----------|----------|----------|----------------|
File             |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
-----------------|----------|----------|----------|----------|----------------|
All files        |        0 |        0 |        0 |        0 |                |
 src             |        0 |        0 |        0 |        0 |                |
  App.vue        |        0 |        0 |        0 |        0 |... 19,20,21,22 |
 src/components  |        0 |        0 |        0 |        0 |                |
  HelloWorld.vue |        0 |        0 |        0 |        0 |... 39,40,41,42 |
-----------------|----------|----------|----------|----------|----------------|```

eddyerburgh:
I'm not sure what to suggest. It looks like Jest has some long standing issues with Windows that I was unaware of. 

All the issues I have looked at suggest that they won't be fixed by the maintainers any time soon 😕

[...]

PR #1073, opened 2017-11-16
refactor: set karma as default

eddyerburgh:
- Set Karma as default
- Run Jest spec files in test/unit, rather than test/unit/specs
https://github.com/vuejs-templates/webpack/issues/1062

### DIFF
--- a/meta.js
+++ b/meta.js
@@ -80,14 +80,14 @@ module.exports = {
       "message": "Pick a test runner",
       "choices": [
         {
-          "name": "Jest",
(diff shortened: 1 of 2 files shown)


LinusBorg:
See #1074 - I think we get Jest running on windows after all :)

robertmain:
FYI: Jest is the only thing that I can get to actually run my tests on windows.

LinusBorg:
It seems the new setup will work, I'll close this for now. We can still revert if we come across other problems.

PR #1082, opened 2017-11-17
Minor refactoring of build scripts

emil14:
Please see the commit messages

### DIFF
--- a/template/build/webpack.base.conf.js
+++ b/template/build/webpack.base.conf.js
@@ -8,6 +8,17 @@ function resolve (dir) {
   return path.join(__dirname, '..', dir)
 }
 
+const lintingRule = {
+  test: /\.(js|vue)$/,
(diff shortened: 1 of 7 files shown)


LinusBorg:
Thanks

template/build/webpack.base.conf.js
```diff
@@ -8,6 +8,17 @@ function resolve (dir) {
   return path.join(__dirname, '..', dir)
 }
 
+const lintingRule = {
+  test: /\.(js|vue)$/,
+  loader: 'eslint-loader',
+  enforce: 'pre',
+  include: [resolve('src'), resolve('test')],
+  options: {
+    formatter: require('eslint-friendly-formatter'),
```
LinusBorg: So this want not a bad idea in hindsight - this will break if the user has decided against using eslint during the project init. Will fix it.

Issue #726, opened 2017-05-25
Headless Chrome for unit tests


frandiox:
Headless Chrome (v59): https://developers.google.com/web/updates/2017/04/headless-chrome
Support in `karma-chrome-launcher`: https://github.com/karma-runner/karma-chrome-launcher/pull/111
Chrome release calendar: https://www.chromium.org/developers/calendar
Windows progress (maybe v60?): https://bugs.chromium.org/p/chromium/issues/detail?id=686608

Hi! Looks like finally, we will be able to use a headless version of Chrome for unit testing. What do you think about using Chrome instead of PhantomJS (or as an extra option)?

I've replaced PhantomJS with this headless Chrome in my local environment ([like this](https://github.com/frandiox/vue-onsenui-webpack/commit/22e109960625f0b5d2535f418eb40a2c2d33a550)) and it works well for now. Web Components and other new APIs can be used in tests :)

corbinu:
Agreed PhantomJS is no longer being developed

LinusBorg:
We will definitely consider this. The question not if, just when.

kazupon:
I checked at [my boilerplate](https://github.com/kazupon/vue-plugin-boilerplate/commit/d2ef3ad674c393802886219ed5cabc6ef8d1761f) too.

It's works! 🚀 

unit test:
<img width="961" alt="screen shot 2017-07-02 at 12 30 37 am" src="https://user-images.githubusercontent.com/72989/27763605-4c1aeb9a-5ec1-11e7-9ff2-8b9b5a380e18.png">

e2e with ngihtwatch:
<img width="493" alt="screen shot 2017-07-02 at 12 37 49 am" src="https://user-images.githubusercontent.com/72989/27763611-72f2c350-5ec1-11e7-8287-ad63b38b9aeb.png">

We should be used Chrome headless instead of PhantomJS. 😉

myst729:
@kazupon The CI server doesn't even have a GUI, is Chrome headless available there?

nickmessing:
@myst729, headless means "without GUI"

frandiox:
@myst729 Just install Google Chrome from your CI recipe. CircleCI example [here](https://github.com/OnsenUI/OnsenUI/blob/master/circle.yml#L26) -- latest `google-chrome-stable` for Linux includes "headless" feature, not sure about other OS.

myst729:
Great! Will definitely give it a try.

nickmessing:
[...]

PR #1098, opened 2017-11-22
Use Chrome Launcher instead of PhantomJS in Karma

Quramy:
[PhantomJS is no longer maintained anymore](https://groups.google.com/forum/#!topic/phantomjs/9aI5d-LDuNE)

### DIFF
--- a/template/test/unit/karma.conf.js
+++ b/template/test/unit/karma.conf.js
@@ -11,8 +11,8 @@ module.exports = function (config) {
     // 1. install corresponding karma launcher
     //    http://karma-runner.github.io/0.13/config/browsers.html
     // 2. add it to the `browsers` array below.
-    browsers: ['PhantomJS'],
-    frameworks: ['mocha', 'sinon-chai', 'phantomjs-shim'],
+    browsers: ['Chrome'],
+    frameworks: ['mocha', 'sinon-chai'],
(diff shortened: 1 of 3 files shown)


sgimeno:
What about using ChromeHeadless as suggested here https://stackoverflow.com/a/41561847? It would fit better for CI setups

ematipico:
Chrome headless is totally better

frandiox:
Related #726 . Not sure if the Karma+Mocha option will be deprecated, though.

template/test/unit/karma.conf.js
```diff
@@ -11,8 +11,8 @@ module.exports = function (config) {
     // 1. install corresponding karma launcher
     //    http://karma-runner.github.io/0.13/config/browsers.html
     // 2. add it to the `browsers` array below.
-    browsers: ['PhantomJS'],
-    frameworks: ['mocha', 'sinon-chai', 'phantomjs-shim'],
+    browsers: ['Chrome'],
```
mdvorscak, line 14: What about ChromeHeadless instead? It's better for CI

lusarz:
I think we need also to "install" Chrome. In my projects I have `puppeteer` added as dev dependency, and line:

    process.env.CHROME_BIN = require('puppeteer').executablePath()

inside `karma.conf.js`.

PR #1102, opened 2017-11-23
* change default host from localhost to 0.0.0.0;

funkyLover:
The server will not be accessed via IP if use `localhost` or `127.0.0.1` as host for webpack-dev-server.


LinusBorg:
Yeah, that's the point - we won't expose the dev-server to the whole netowrk by default.

What I would accept and merge would be a comment in the config file mentioning that `0.0.0.0` is required for outside access of the dev-server.

PR #1103, opened 2017-11-23
Display actual host when showing friendly server listening notification

### LINKED ISSUE #1101: If HOST is provided through environment, friendly notification won't display the actual one
There is this: https://github.com/vuejs-templates/webpack/blob/7cf7d4134607348f296287b661dfa8764880c605/template/build/webpack.dev.conf.js#L68

If host is provided through environment, this won't get updated and will still display the value found in config, not taking in mind that it could be changed here: https://github.com/vuejs-templates/webpack/blob/7cf7d4134607348f296287b661dfa8764880c605/template/build/webpack.dev.conf.js#L24

It's a simple fix, though: 
```
messages: [`Your application is running here: http://${devWebpackConfig.devServer.host}:${port}`]
```

I won't bother with expected/actual behaviour, because the code lines should be simple to follow.

---

Posted in response to https://github.com/vuejs-templates/webpack/issues/1099#issuecomment-346556740, because if the switch was made to use internal portfinder of `webpack-dev-server` this message will most definitely need a check-up.

joltmode:
Fixes #1101.

### DIFF
--- a/template/build/webpack.dev.conf.js
+++ b/template/build/webpack.dev.conf.js
@@ -65,7 +65,7 @@ module.exports = new Promise((resolve, reject) => {
       // Add FriendlyErrorsPlugin
       devWebpackConfig.plugins.push(new FriendlyErrorsPlugin({
         compilationSuccessInfo: {
-          messages: [`Your application is running here: http://${config.dev.host}:${port}`],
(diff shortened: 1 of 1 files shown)


LinusBorg:
Thanks!

PR #1115, opened 2017-11-27
Add postcss-url to match postcss-import (fix #1106)

### LINKED ISSUE #1106: postcss import need works with postcss url
Hi, I got error in last release

my css:

```css
@font-face {
  font-family: "iconfont";
  src: url('iconfont.eot?t=1510823274364'); /* IE9*/
  /* ... */
}
```

it is solved when I add `"postcss-url": {},` in `.postcssrc.js`

SirM2z:
fix like this [postcss import need works with postcss url #1106](https://github.com/vuejs-templates/webpack/issues/1106)

### DIFF
--- a/template/.postcssrc.js
+++ b/template/.postcssrc.js
@@ -2,8 +2,9 @@
 
 module.exports = {
   "plugins": {
-    // to edit target browsers: use "browserslist" field in package.json
     "postcss-import": {},
+    "postcss-url": {},
+    // to edit target browsers: use "browserslist" field in package.json
     "autoprefixer": {}
   }
 }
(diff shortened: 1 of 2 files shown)


template/.postcssrc.js
```diff
@@ -4,6 +4,7 @@ module.exports = {
   "plugins": {
     // to edit target browsers: use "browserslist" field in package.json
```
LinusBorg: could you move this line so it's right above `autoprefixer`? That's where it belongs. Thanks! :)

SirM2z: I have already modified it.

LinusBorg:
Seems to solve issues with reltive paths in CSS libs as well, so we need this (see #1199)

PR #1118, opened 2017-11-27
Update project creation: sort dependencies in package.json

jakub300:
Additionally update devDependencies in package.json: conditional dependencies are at the top, all other are below them.


LinusBorg:
Awesome work, thanks!

PR #1119, opened 2017-11-28
Switch to uglifyjs-webpack-plugin

sudo-suhas:
### Summary

This switches to [`uglifyjs-webpack-plugin`](https://github.com/webpack-contrib/uglifyjs-webpack-plugin) for minification which uses [UglifyJS v3 (`uglify-es`)](https://npmjs.com/package/uglify-es) to minify javascript. 

### Why is this required

Right now, we use `babel-preset-env` with some browser versions as the target. This dynamic config may break minification because currently the uglify plugin shipped in webpack 2 doesn't support ES6+. For example, if I change my `target.browsers` to `["last 2 Chrome versions", "last 2 Firefox versions"]` it will break uglify. Since this template is meant to be customised by developers, it would be better to have a more robust method for minification.

Also see discussion in vuejs-templates/pwa#44. If this PR is approved here, I'll make a PR to the `pwa` template as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ The development server will run on port 8080 by default. If that port is already
   - Source maps
 
 - `npm run build`: Production ready build.
-  - JavaScript minified with [UglifyJS](https://github.com/mishoo/UglifyJS2).
+  - JavaScript minified with [UglifyJS v3](https://github.com/mishoo/UglifyJS2/tree/harmony).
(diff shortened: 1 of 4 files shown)


LinusBorg:
How does this compare to babel-minify? Any experience with both?

sudo-suhas:
I had tried `babel-minify` a few months back and the bundle size was slightly larger.

LinusBorg:
Thanks

PR #1120, opened 2017-11-28
babel-register use only for e2e tests

PowerfulPony:
Single require here template/test/e2e/nightwatch.conf.js


LinusBorg:
What would be the advantage of this?

PowerfulPony:
If you do not use e2e, the unused dependency will not be loaded. When studying the project, there is no question of why this dependence is necessary.

LinusBorg:
> there is no question of why this dependence is necessary.

Then maybe I'm stupid. Enlighten me.

LinusBorg:
Fixed the merge conflict created by #1118

PR #1126, opened 2017-11-30
Prevent third-party libraries from repeatedly importing vue

linjiajian999:
一些第三方库在引入vue.js到时候，使用如下配置
Some third-party libraries use this configuration when use vue.js
```js
...
externals:  {
    vue: {
      commonjs: 'vue',
      commonjs2: 'vue',
      amd: 'vue',
      root: 'Vue',
      var: 'Vue'
    }
  }
...
```
这使得本地项目在引入第三方库并且打包的时候重新引入vue
此时，打包第三方库时如（vue-material）,第三方库会引入（vue-material/node_modules/vue/dist/vue.esm.js）

It wiil import vue into packaged duplicate

image

因此，修改此处别名
Therefore, change the 'vue' of alias to 'path.resolve(__dirname, '../node_modules/vue/dist/vue.esm.js'),'

```
// build/webpack.base.conf.js
resolve: {
    extensions: ['.js', '.vue', '.json'],
    alias: {
       'vue$': path.resolve(__dirname, '../node_modules/vue/dist/vue.esm.js'),
      '@': resolve('src')
    }
  },
```
使得webpack打包的时候，始终是打包到同一个vuejs
Then, it will import vue once time
image

### DIFF
--- a/template/build/webpack.base.conf.js
+++ b/template/build/webpack.base.conf.js
@@ -24,7 +24,7 @@ module.exports = {
     extensions: ['.js', '.vue', '.json'],
     alias: {
       {{#if_eq build "standalone"}}
-      'vue$': 'vue/dist/vue.esm.js',
+      'vue$': path.resolve(__dirname, '../node_modules/vue/dist/vue.esm.js'),
       {{/if_eq}}
       '@': resolve('src'),
     }


LinusBorg:
Wow, good find!

Will check this out this weekend.

LinusBorg:
@chrisvfritz This would probably also be relevant for the docs.

chrisvfritz:
I don't think this is actually a good idea. The real problem is that vue-material has listed Vue as a dependency, which means they specifically want to package their own version of Vue into the library. This is most likely a mistake and what they really want is to list Vue as a [peer dependency](https://docs.npmjs.com/files/package.json#peerdependencies). That's the correct way to avoid bundling the same package twice.

linjiajian999:
@chrisvfritz get! I think you are right! Thanks for your check out this pr.

PR #1130, opened 2017-12-01
use postcss default in vue-loader options

teazean:
It seems like loader in vue-loader's options contains no postcss, and but it should have.

### DIFF
--- a/template/build/vue-loader.conf.js
+++ b/template/build/vue-loader.conf.js
@@ -10,7 +10,8 @@ const sourceMapEnabled = isProduction
 module.exports = {
   loaders: utils.cssLoaders({
     sourceMap: sourceMapEnabled,
-    extract: isProduction
+    extract: isProduction,
+    usePostCSS: true
   }),
   cssSourceMap: sourceMapEnabled,
   transformToRequire: {


LinusBorg:
Thanks, but that's not necessary.  vue-loader runs everything through postcss anyway.

That's why that option is there in the first place: so we add postcss-loader only for the style-loaders for `.css` files etc, and *not* for `.vue`

teazean:
Not exactly as you said.
Firstly in https://github.com/vuejs-templates/webpack/blob/develop/template/build/vue-loader.conf.js#L10

```
 loaders: utils.cssLoaders({
    sourceMap: sourceMapEnabled,
    extract: isProduction
  }),
```
There's no usePostcss option.

And in https://github.com/vuejs-templates/webpack/blob/develop/template/build/utils.js#L34

```
  const loaders = options.usePostCSS ? [cssLoader, postcssLoader] : [cssLoader]

  // others
  // https://vue-loader.vuejs.org/en/configurations/extract-css.html
  return {
    css: generateLoaders(),
    postcss: generateLoaders(),
    less: generateLoaders('less'),
    sass: generateLoaders('sass', { indentedSyntax: true }),
    scss: generateLoaders('sass'),
    stylus: generateLoaders('stylus'),
    styl: generateLoaders('stylus')
  }
}
```

The default does not apply to postcss-loader.

LinusBorg:
You are misunderstanding what I wrote: 

All css is run through postcss *inside of vue-loader*. Always. Without any config necessary. 

It's he docs, just take a look.

teazean:
Yep, thank you for answering.

https://vue-loader.vuejs.org/en/features/postcss.html

**Any CSS output processed by vue-loader is piped through PostCSS for scoped CSS rewriting. **

PR #1131, opened 2017-12-01
Add option to allow remote access on webpack-dev-server

Toilal:
Set this option to allow remote access to webpack dev server (avoid Invalid Header error)

### DIFF
--- a/template/build/webpack.dev.conf.js
+++ b/template/build/webpack.dev.conf.js
@@ -26,6 +26,7 @@ const devWebpackConfig = merge(baseWebpackConfig, {
     compress: true,
     host: HOST || config.dev.host,
     port: PORT || config.dev.port,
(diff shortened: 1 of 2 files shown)


LinusBorg:
Hey @Toilal 

Thanks for the PR. a quick question: host `0.0.0.0` is enough to access from within the same network, right? Like from a mobile phone in my Wifi. Or would that also require this additional option?

Toilal:
Setting host to `0.0.0.0` is NOT enough, because dev server check the HTTP `host` header actually match what's is defined in `host` parameter. If it doesn't match, it shows an empty HTML page with "Invalid Host Header".

It can be useful for mobile access from desktop or Dockerized yarn installations.

LinusBorg:
Thanks, understood.

LinusBorg:
(close and re-open to trigger CI)

PR #1135, opened 2017-12-03
webpack.test.conf.js is not needed in jest and e2e

pwang2:
As titled.


LinusBorg:
That's correct, good catch, thanks.

PR #1160, opened 2017-12-12
Add more integration tests (expect scripts)

Toilal:
This adds more integration tests with [expect](https://linux.die.net/man/1/expect) scripts and some bash scripts.

I run this kind of tests on my [fork](https://github.com/Toilal/vue-webpack-template) since months and it does the job. I'm using TravisCI on my fork, but it should work with CircleCI too.

This check for many answers series (defined in expect scripts) if generation, lint, tests and build returns `0` exit code.


Toilal:
For some reason, those tests fails when using the brand new automated "npm install", so expect script are written to answer "No" to the last question, and bash script runs the commands manually.

LinusBorg:
Hey, 

I have been working on improving the CI tests for some time, and took a very different approach - one that doesn't rely on bash so windows users can run them too.

Please have a look at #1218

PR #1161, opened 2017-12-12
Fix babelrc testing env

Toilal:
Small mistake that can make tests failing in some cases

### DIFF
--- a/docs/env.md
+++ b/docs/env.md
@@ -20,7 +20,7 @@ module.exports = merge(prodEnv, {
 
 // config/test.env.js
 module.exports = merge(devEnv, {
-  NODE_ENV: '"testing"'
+  NODE_ENV: '"test"'
 })
 ```
 
@@ -36,7 +36,7 @@ So, the environment variables are:
     - DEBUG_MODE = true,
     - API_KEY    = '...'
 - Testing
-    - NODE_ENV   = 'testing',
+    - NODE_ENV   = 'test',
     - DEBUG_MODE = true,
     - API_KEY    = '...'
 
--- a/template/config/test.env.js
+++ b/template/config/test.env.js
@@ -3,5 +3,5 @@ const merge = require('webpack-merge')
 const devEnv = require('./dev.env')
 
 module.exports = merge(devEnv, {
-  NODE_ENV: '"testing"'
+  NODE_ENV: '"test"'
 })


LinusBorg:
Please explain in what cases that can make tests fail. This is a pretty complicated array of possible setups we have to cater to (with/whithout e2e, with/without unit tests, unit test with jest or with karma...) so it's important to know what you think you are fixing, in order to  which other side effects this might have.

Toilal:
After some additions to support vue-class-components with ES2015 scripts, I experienced some errors when running e2e tests. babel plugins declared in `test` env of `.babelrc` were not loaded.

I then noticed that `NODE_ENV=testing` was declared in the codebase

https://github.com/vuejs-templates/webpack/blob/cc89353be444eee88e7e6480bf186a3cf5bd439f/template/test/e2e/runner.js#L2

https://github.com/vuejs-templates/webpack/blob/17ed63b1b3a0eaaebd3f593c08c32107a7cb7e01/template/config/test.env.js#L6

but it was inconsistent with BABEL_ENV="test" declared in package.json scripts for unit tests, and inconsistent with .babelrc env entry

With those changes, it seems to fix my issue. Don't know why the CI fails, it seems unrelated to those changes.

LinusBorg:
> With those changes, it seems to fix my issue. Don't know why the CI fails, it seems unrelated to those changes.

You didn't change the BABEL_ENV for jest, and jest sets `NODE_ENV` to `test` unless specified otherwise - which in turn is the variable babel falls back to if BABEL_ENV in't specified either.

> {{#if_eq runner "jest"}}
>   "unit": "jest --config test/unit/jest.conf.js --coverage",
 > {{/if_eq}}

So your change breaks tests with jest - but the unit command for jest could be adjusted accordingly to fix this.

However, I would like to learn more about what additional plugins you needed for e2e tests. e2e tests are starting a webpack-dev-server with a production config, so I don't see what special plugins would be required for babel during e2e tests. after all, the dev-server is supposed to behave just like a production build.

I'm not saying there can'T be anything wrong with the setup we currently have, but we might have to change something else entirely depending on the source of the issues you faced.

Toilal:
Ok I see. 

Maybe the `testing` occurences could be replaced by `test` then ? If it's convention for most testing tools, it would make things less confusing (I didn't know about jest adding node_env="test" automatically).

Long explanation of my particular case:

I also have a "Library" option in my fork that split sources in 2 directories : "app", and "src". 

"src" contains sources of the library that will be included in the production build, and "app" contains a UI that can be use to build a docs/demo website from the library, helps the development of the library by providing a UI on top of it.

Production build only bundle the entrypoint from "src", but E2E tests needs to run against the "app", so I have made a change to run E2E tests against the dev configuration.

But in fact, I think I should create 2 production builds, one for the lib, and one for the "app", and run E2E tests against the production build of the "app".

Feel free to close this one If you don't want to replace the `testing` occurences to `test`, else I can update the pull request.

Toilal:
I have update the Pull Request with the proposal of renaming "testing" to "test". I have removed cross-env dependency as it's not used anymore, because NODE_ENV="test" is defined by webpack configuration.

LinusBorg:
I will close this PR since it triggered me to work on #1163 which will bring some changes to the area affected by this PR and others.

PR #1165, opened 2017-12-13
Remove useless code when lint is not chosen

Toilal:


template/build/webpack.base.conf.js
```diff
@@ -8,7 +8,7 @@ function resolve (dir) {
   return path.join(__dirname, '..', dir)
 }
 
-const createLintingRule = () => ({
+{{#lint}}const createLintingRule = () => ({
```
LinusBorg: There's a call to that function further below in the rules section, that should be wrapped in the same handlebars check.

Toilal: It's already wrapped

LinusBorg: Right, my mistake. Thanks.

PR #1171, opened 2017-12-14
Gracefully handle npm install error gracefully (close #1168)

### LINKED ISSUE #1168: Auto-Install fails on some OS/node combinations
### The problem:

For the original bug report, see https://github.com/vuejs/vue-cli/issues/681

### Immediate solutions/workarounds.

One thing we can immediately improve is the error handling. This error is not "fatal" - all people have to do is run `npm install` themselves. So I will add a nice error message explaining that to users.

### so what's the Problem, really?

As to the original source of the problem, I'm a bit dumbfounded for the moment.

> When you then elect to have scaffolding restore your packages (whether through npm or yarn), you are not in the project directory, but above it, so the package install/restore command fails

That would be weird - being in the wrong directory should lead to an error on OSX as well, but doesn't (for me). 

I also think I set the cwd for the npm call to the right directory here: 

https://github.com/vuejs-templates/webpack/blob/develop/meta.js#L155-L158

LinusBorg:
### The problem

See #1168 

### what does this PR do

When the user has chosen to let the template automatically install dependencies, and that call to `npm install` fails, We now display a usefull error message with instructions on how to proceed.

bildschirmfoto 2017-12-14 um 12 57 04

### Is that a perfect solution?

Not really, but it helps to proceed even if autoInstall fails.

If someone for whom this process fails would be willing to investigate, that would be much apprechiated. Since I can't reproduce the error from #1168, I don't know how to debug this.

### To test this PR:

```
vue init webpack#fix_installerror_1168 
```

I'm mentioning some folks who have previously reported this error with the request to test this workaround/solution on their systems:

* @xiejiamiao
* @KyleLehtinenDev
* @IcedKnight
* @songhongjiang
* @lambdakris

Thanks!

### DIFF
--- a/meta.js
+++ b/meta.js
@@ -155,9 +155,9 @@ module.exports = {
     const cwd = path.join(process.cwd(), data.inPlace ? '' : data.destDirName)
 
     if (data.autoInstall) {
-      installDependencies(cwd, data.autoInstall, green)
+      installDependencies(cwd, data.autoInstall, chalk)
         .then(() => {
-          return runLintFix(cwd, data, green)
+          return runLintFix(cwd, data, chalk)
         })
         .then(() => {
           printMessage(data, green)
--- a/meta.js
+++ b/meta.js
@@ -155,7 +155,7 @@ module.exports = {
     const cwd = path.join(process.cwd(), data.inPlace ? '' : data.destDirName)
 
     if (data.autoInstall) {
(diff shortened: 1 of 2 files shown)


LinusBorg:
Damn, it seem #1170 might fix the underlying reason, and it was easy so find. I will close this PR, and push another one rebased on that fix.

PR #1172, opened 2017-12-14
Handle npm install error gracefully

LinusBorg:
### The problem

See #1168 - it ssems to be fixed by #1170 already, but imrpoving the error handling in case something breaks might still be a good idea.

Thougths?

### What does this PR add/change?

When the user has chosen to let the template automatically install dependencies, and that call to `npm install` fails, We now display a usefull error message with instructions on how to proceed.

bildschirmfoto 2017-12-14 um 12 57 04

### Is that a perfect solution?

Not really, but it helps to proceed even if autoInstall fails.

If someone for whom this process fails would be willing to investigate, that would be much apprechiated. Since I can't reproduce the error from #1168, I don't know how to debug this.

### To test this PR:

```
vue init webpack#fix_installerror_1168 
```

Thanks!

### DIFF
--- a/meta.js
+++ b/meta.js
@@ -155,9 +155,9 @@ module.exports = {
     const cwd = path.join(process.cwd(), data.inPlace ? '' : data.destDirName)
 
     if (data.autoInstall) {
-      installDependencies(cwd, data.autoInstall, green)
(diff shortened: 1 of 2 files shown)


LinusBorg:
Closing this as I think we don't need it after all.

PR #1181, opened 2017-12-17
Add proxy setting for karma test.

matyapiro31:
If not, web server returns 404 if it is referring files in static/

### DIFF
--- a/template/test/unit/karma.conf.js
+++ b/template/test/unit/karma.conf.js
@@ -18,6 +18,9 @@ module.exports = function (config) {
     preprocessors: {
       './index.js': ['webpack', 'sourcemap']
     },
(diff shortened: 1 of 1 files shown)


LinusBorg:
How do static files interfer with your unit tests?

matyapiro31:
@LinusBorg There are warning when vue refers files in /static/.
Also in my project, some files that is output to /static when building.

LinusBorg:
> There are warning when vue refers files in /static/.

What kind of warnings? I still don't see what warnings would appear in unit tests in these cases.

LinusBorg:
Closing for inactivity. It want explains what problem this PR solves, so I don't see a reason to merge it.

Issue #775, opened 2017-06-28
Add bail option to production webpack config


duebbert:
webpack happily ignores build errors which is ok for development but when creating production files it should fail. 

The configuration option [`bail`](https://webpack.github.io/docs/configuration.html#bail) should be used for this and results in a build to throw an error when a problem is encountered instead of continuing.

Proposal: add `bail: true` to the configuration in `build/webpack.prod.conf.js`

LinusBorg:
I think this makes sense. Sorry for overlooking it for so long.

PR #1189, opened 2017-12-18
Enable bail option to webpack prod configuration

afontcu:
Hi! This should solve https://github.com/vuejs-templates/webpack/issues/775.

Thanks for ur work!

### DIFF
--- a/template/build/webpack.prod.conf.js
+++ b/template/build/webpack.prod.conf.js
@@ -29,6 +29,9 @@ const webpackConfig = merge(baseWebpackConfig, {
     filename: utils.assetsPath('js/[name].[chunkhash].js'),
     chunkFilename: utils.assetsPath('js/[id].[chunkhash].js')
   },
+  // Exit bundling process on the first error instead of tolerating it:
+  // https://webpack.js.org/configuration/other-options/#bail
+  bail: true,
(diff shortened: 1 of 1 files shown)


LinusBorg:
Will make a quick test run with this to see the behaviour when it fails, then I'll merge. Thanks.

LinusBorg:
So I just tested this. The problem I'm seeing with this is that the error message generated is not as good as it is without that option, at least in case of a missing module which I used as a test case.

### With `bail: true`:

bildschirmfoto 2018-01-04 um 18 44 19

The error states that a module is missing, but yu don'T really learn *in what file* that error happened.

### With the current config:

bildschirmfoto 2018-01-04 um 18 45 53

Now that's much more helpful, isn't it?

Maybe there's a way to get these kinds of error messages with `bail: true` as well, but I don't know of any.

If you (or someone else) can come up with a solution for that, I'm willing to consider switching to `bail: true`, but I don't think we should accept inferior error messages for a quicker build cancellation when an error is encountered.

afontcu:
You are completely right, I'd rather stick with the current config. I'm afraid I don't know how to improve bail error output, so feel free to close the PR if you think it's been waiting for help long enough.

Thanks!

PR #1201, opened 2017-12-25
downgrade uglifyjs-webpack-plugin to v0.4

thinhbuzz:
because v1.0 using for webpack 4.0:

from [uglifyjs-webpack-plugin](https://github.com/webpack-contrib/uglifyjs-webpack-plugin):

> ℹ️  `webpack =< v3.0.0` currently contains [`v0.4.6`](https://github.com/webpack-contrib/uglifyjs-webpack-plugin/tree/version-0.4) of this plugin under `webpack.optimize.UglifyJsPlugin` as an alias. For usage of the latest version (`v1.0.0`), please follow the instructions below. Aliasing `v1.0.0` as `webpack.optimize.UglifyJsPlugin` is scheduled for `webpack v4.0.0`

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -107,7 +107,7 @@
     "optimize-css-assets-webpack-plugin": "^3.2.0",
     "ora": "^1.2.0",
     "rimraf": "^2.6.0",
-    "uglifyjs-webpack-plugin": "^1.1.1",
+    "uglifyjs-webpack-plugin": "^0.4.6",
     "url-loader": "^0.5.8",
     "vue-loader": "^13.3.0",
     "vue-style-loader": "^3.0.1",


LinusBorg:
That quote alone is no reason to not use this plugin. It is meant to work with webpack 3 just as well - it just won't be included *into* webpack before 4.0.0

But maybe we have to temporarily downgrade because of https://github.com/webpack-contrib/uglifyjs-webpack-plugin/issues/197

LinusBorg:
Note: downgrading the version won't be enough, as plugin options work differenlty for 0.4 and 1.0.0:

https://github.com/vuejs-templates/webpack/pull/1119

thinhbuzz:
why do you want to ignore offcial note of plugin?

LinusBorg:
I didn't ignore it. I explained to you how I think you misunderstand it.

That plugin has a `1.0.0` release - that means it's stable. The note only says *"we will *inlcude* this in the webpack default plugins starzing with 4.0.0"*. It does **not** say *"don't use this plugin, it's not ready yet"*

LinusBorg:
Problem was fixed with v1.16 of the plugin, so we can close this.

PR #1202, opened 2017-12-26
Load, render template version using helper

sudo-suhas:
This adds a helper to `meta.js` so that the template version in `template/config/index.js` does not need manual maintenance and is instead loaded from `package.json>version`.

### DIFF
--- a/meta.js
+++ b/meta.js
@@ -6,16 +6,22 @@ const {
   runLintFix,
   printMessage,
 } = require('./utils')
+const pkg = require('./package.json')
+
+const templateVersion = pkg.version
 
 module.exports = {
   helpers: {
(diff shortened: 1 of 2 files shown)


LinusBorg:
@sudo-suhas That actually doesn't work for me, I just get this:

bildschirmfoto 2017-12-31 um 14 51 05

it correctly reads the version from pkg.json, but apperantly the helper returns nothing for me :/

sudo-suhas:
My bad. I had tested it but must have inadvertently introduced the mistake while rebasing and removing conflicts. Very sorry.

LinusBorg:
Don't worry, thanks for the continuing contributions :)

PR #1218, opened 2018-01-04
Improve template CI tests

LinusBorg:
This PR completely changes the CI testing setup **for the template** - not projects generated with it, so id doesn't affect users in any way.

### Changes

* introduces ability to test different answer scenarios.
* no more `test.sh`, simply set an env variable to roll out a template with a defined set of answers.
* uses circleCI 2.0 now, for improved performance. all scenarios (currently 3) are tested in parallel.

#### Minimal

Answers "no", to all choices, so the absolutely basic setup.

#### Full

Answers "yes" to all choices, using the default value where possible. so it runs unit tests with jest and eslint with the "standard" config

#### Full-karma-airbnb

Like "Full", but uses karma instead of jest and airbnb eslint config instead of standard.

 ## Open TODOs:

- [ ] add npm script to run tests locally
- [ ] remove `test.sh` file


Toilal:
It's actually far better than expect scripts ! Thanks :)

It's a shame inquirer.js doesn't support some kind of automation, the `isNotTest` variable seems like a hack, but it's still better than expect script that are just *** to build and maintain.

Toilal:
(Don't forget to squash commits :rofl: :lollipop:)

LinusBorg:
Oh, I always squash PRs, but this one needs it badly - I fought with circleci config forever 😂😂

codetheorist:
Could we not do away with the shell script all together now. If all of the answers are in scenarios, there's technically no need for the yes command.

Would there be a way to run all scenarios with one command?

LinusBorg:
Oh right, the fact that the test.sh file is still there is an oversight. 

Abs  should add am no script to run those tests in sequence. 

For CI we run them in parallel with circleci commands, that's why in haven't added a script yet.

PR #1221, opened 2018-01-05
Add HardSourceWebpackPlugin to improve compiling time

soanvig:
**Problem**
Running dev server with `npm run dev` takes about 40s to compile, each time. Same thing with production builds.

**Solution**
Using HardSourceWebpackPlugin improves caching. Therefore launching dev server first time takes 40s, launching it second time takes 1s.

Source:
https://medium.com/ottofellercom/0-100-in-two-seconds-speed-up-webpack-465de691ed4a

### DIFF
--- a/template/build/webpack.base.conf.js
+++ b/template/build/webpack.base.conf.js
@@ -3,6 +3,7 @@ const path = require('path')
 const utils = require('./utils')
 const config = require('../config')
 const vueLoaderConfig = require('./vue-loader.conf')
+const HardSourceWebpackPlugin = require('hard-source-webpack-plugin')
 
 function resolve (dir) {
   return path.join(__dirname, '..', dir)
@@ -24,6 +25,9 @@ module.exports = {
   entry: {
     app: './src/main.js'
   },
+  plugins: [
+    new HardSourceWebpackPlugin()
+  ],
   output: {
     path: config.build.assetsRoot,
     filename: '[name].js',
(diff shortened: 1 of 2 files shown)


heinzdmx:
My 2-cents on this: 

`hard-source-webpack-plugin` seems very promising, but looking at https://github.com/webpack/webpack/issues/250#issuecomment-347811223 and the comments on it, it appears that it relies heavily on webpack internals which means that even a semver-minor change could result in breaking `hard-source-webpack-plugin`
I woud therefore caution about adding it to this template and instead suggest that it might be added to a readme / added as commented out code.

soanvig:
@heinzdmx , theoretically, yes, but the webpack template isn't up-to-date with official webpack releases. AFAIK it is always few patches behind official release. By this time `hard-source` developers should make their release.

But You made your point, I'm almost conviced, that there should be note in README. Webpack compilation is really slow at that moment. Some solution is required.

LinusBorg:
I'll have to side with @heinzdmx on this one. The plugin doesn'T seem to be of the kind you should be using without knowing what it does or what risks you are taking by using it.

Also, this template does not attempt to provide a setup that will work perfectly for gigantic projects - those will always need individual optimization. 

So I would argue that this is a kind of plugin that users should add at their own risk, once they actually have a performance problem they need to solve.

sergiocastrovale:
@LinusBorg what kind of optimizations would you suggest? I have a fairly large project that takes 8-10s to rebuild on `dev` whenever I save anything.

LinusBorg:
Please consider using forum.vuejs.org for project-related questions. This issue is closed, and github issues in general are not supports tickets.

Issue #1226, opened 2018-01-06
Add easy-deploy option when installing the template using Vue CLI


codetheorist:
Add an option to the template installer that allows the user to add easy-deployment for GitHub pages.

If the user opts-in to this option, they will then be presented with 2 further prompts for:

- Branch Name
- Base URL (for project pages)

Once these details are entered, the will then simply have to run npm run deploy to deploy a built site to the correct branch of the repository.

lubien:
Hi there. 

This is a feature I find doing myself a lot and I'd love to have here on this template.

Mind if I ask if there'll be any progress in this issue or, if that's the case, could I take over it?

Edit: I did some digging. Seems that @vue/cli@3 is the new way. I'll look into that.

PR #1230, opened 2018-01-06
Issue #1226: Add easy-deploy option when installing the template using Vue CLI

codetheorist:


codetheorist:
The tests failed on this however further investigation revealed issues with a previously merged commit.

A working demo built using this Webpack template is available on GitHub [here](https://codetheorist.github.io/webpack-deploy).

The actual template that is generated is also available on the master branch for the repo [here](https://github.com/codetheorist/webpack-deploy).

LinusBorg:
We generally have some issues with CI tests, which is why I submitted #1218.

But you submitted 2 versions of this PR, can the old one be closed?

template/package.json
```diff
@@ -103,6 +106,9 @@
     "postcss-import": "^11.0.0",
     "postcss-loader": "^2.0.8",
     "postcss-url": "^7.2.1",
+    {{#deploy}}
+    "push-dir": "^0.4.1",
```
LinusBorg: There's a package that makes this even easier imho: gh-pages

Have you considered it?

codetheorist:
Yeah, I did leave a comment on the old PR with regards to closing it for me please.

LinusBorg:
Closed it.. by you can close your own PRs yourself I think ;)

template/package.json
codetheorist: I did consider the gh-pages package, but when checking out both of the packages documentation, gh-pages just seems to make it more complicated than it should be.

With the push-dir package, any flags can simply be added to the package.json file, such as --allow-unclean and --force, if required by the end-user.

codetheorist:
I did try to close, but I couldn't for the life of me find how to do it. It was probably right in front of me :)

Issue #1217, opened 2018-01-03
config.dev.https not read into devWebpackConfig.devServer


bird-cage:
**The BUG**
I was unable to view my dev instance with https even though I'd set config.dev.https: true.  
It did work when I added --https to the package.json scripts.dev cli options
After looking around I seen that in ./build/webpack.dev.config.js where the Promise for the server is defined, config.dev.https was not being read into const: devWebpackConfig.devServer

**The FIX**
In ./build/webpack.dev.config.js under the lines:
    host: process.env.HOST || config.dev.host,
    port: process.env.PORT || config.dev.port,
I added:
    https: config.dev.https ? config.dev.https: false,

**Now all works for me as expected**
This should be fixed ;-)

**P.S. as an additional tip** 
for somebody tackling using https with the webpack dev proxy server like me, be sure to not only set:
proxyTable.target = 'https://yourOtherServersDomain[:andPortIfRequired]'
BUT ALSO
proxyTable.secure = false _This may only be necessary if it's an self-signing key, which it is in my case_

**The errors I seen which could help somebody find this when:**
_devServer.https not set_
Browser: 
This site can’t be reached

_devServer is running with https but proxyTable.secure != false_
Terminal:
[HPM] Error occurred while trying to proxy request /webshop/product/lottonormal from xxxx:8080 to https://xxxxxx.dev.local:24910 (UNABLE_TO_VERIFY_LEAF_SIGNATURE) (https://nodejs.org/api/errors.html#errors_common_system_errors)
Chrome network dev Tab:
Headers status code 500

bird-cage:
PPS, I've never done it before on an open-source repo, but I will do a pull-request to fix this as suggested above within the next couple of days

LinusBorg:
Hey, thanks for wanting to contribute, apprechiated! If you need help, just ask. 

As a small note, I would not qualify this as as "bug" because `/config/index.js` is not meant to dynamically add any dev server options to the webpack config. It's just a static subset of options that we extracted out of webpack config for easier access. Adding options to this file will do nothing. 

But adding https support can be considered a worthwhile enhancement, so go ahead :)

PR #1233, opened 2018-01-07
Mapped config.dev.https to devWebpackConfig.devServer (#1217)

bird-cage:
@LinusBorg 
The changes are similar as described at https://github.com/vuejs-templates/webpack/issues/1217, with the addition that in the notification message http|https are show as appropriate.

It took me a while to realize I had to $ vue init bird-cage/webpack **#dev** (I'm new to this ;-), but then it worked as expected.

What do you think?

What I also see, but can't figure out how to improve (assuming you would see this as an enhancement) is if for example the port is set in package.json scripts dev: .... --port 8181 the dev server will use this port instead of that in config/index.js, but the notification message still reads ...:8080 or whatever is set in config/index.js

### DIFF
--- a/template/build/webpack.dev.conf.js
+++ b/template/build/webpack.dev.conf.js
@@ -31,6 +31,7 @@ const devWebpackConfig = merge(baseWebpackConfig, {
     hot: true,
     contentBase: false, // since we use CopyWebpackPlugin.
     compress: true,
+    https: config.dev.https,
     host: HOST || config.dev.host,
     port: PORT || config.dev.port,
     open: config.dev.autoOpenBrowser,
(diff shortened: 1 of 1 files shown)


template/build/webpack.dev.conf.js
```diff
@@ -31,6 +31,7 @@ const devWebpackConfig = merge(baseWebpackConfig, {
     hot: true,
     contentBase: false, // since we use CopyWebpackPlugin.
     compress: true,
+    https: config.dev.https,
```
LinusBorg: Thanks! We still have to add the actual property to `/config/index.js` file so people can know about the option's existance without studying the webpack files.

PR #797, opened 2017-07-13
Minimal TypeScript support option added

Igogrek:
This PR adds TypeScript language option to webpack template.

This is basically the same idea as in: https://github.com/vuejs-templates/webpack/pull/781

The main difference is that I've tried to find minimal possible configuration and code changes.
So in this case there's a metalsmith plugin that renames .js files from /src to .ts if typescript option is enabled.
This way there's no need to maintain both .js and .ts sources.
Checked with both TS enabled and disabled.

PR #1238, opened 2018-01-10
Typescript capablility

ffxsam:
Modified webpack template to allow creation of TS projects (including TSLint)

### DIFF
--- a/template/package.json
+++ b/template/package.json
@@ -87,7 +87,7 @@
     "tslint": "^5.8.0",
     "tslint-loader": "^3.5.3",
     {{/tslint}}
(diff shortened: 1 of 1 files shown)


ffxsam:
It's a little opinionated as it includes some basic tslint rules. I can remove those if it's an issue.

LinusBorg:
Hey, Thanks for your work on this!

However, we already have a longstanding PR open with #797, which I intend to review and merge after #1218

So I doub that this PR will make it. However I would appreachiate if you would comment on the work in #797 and add feedback from your experience

PR #1251, opened 2018-01-18
Update webpack.dev.conf.js

epver:
update webpack.dev.conf.js, for disableHostCheck


LinusBorg:
From the webpack docs:

> When set to true this option bypasses host checking. THIS IS NOT RECOMMENDED as apps that do not check the host are vulnerable to DNS rebinding attacks.

I don't think we should add convenience config options that aren't recommended. If users need them, they should add them themselves and know what they are doing.

Issue #535, opened 2017-02-20
Module build failed: Error: "extract-text-webpack-plugin" loader is used without the corresponding plugin


ghost:
Thank you for an awesome template. 
I have run into this error with Windows 10, node 7.5.0 and npm 4.1.2. 
This error is with the boilerplate code unmodified  and running with NODE_ENV=production. 
There is no error with NODE_ENV=testing or development.
_________________________________________________________________________________________
``` 
ERROR  Failed to compile with 2 errors   

 error  in ./src/App.vue

Module build failed: Error: "extract-text-webpack-plugin" loader is used without the corresponding plugin, refer to https://github.com/webpack/extract-text-webpack-plugin for the usage example
    at Object.module.exports.pitch (C:\iTurbineIO\Projects\vue-webpack\node_modules\extract-text-webpack-plugin\loader.js:27:9)

 @ ./src/App.vue 3:0-335
 @ ./src/main.js
 @ multi ./build/dev-client ./src/main.js

 error  in ./src/components/Hello.vue

Module build failed: Error: "extract-text-webpack-plugin" loader is used without the corresponding plugin, refer to https://github.com/webpack/extract-text-webpack-plugin for the usage example
    at Object.module.exports.pitch (C:\iTurbineIO\Projects\vue-webpack\node_modules\extract-text-webpack-plugin\loader.js:27:9)

 @ ./src/components/Hello.vue 3:0-345
 @ ./src/router/index.js
 @ ./src/main.js
 @ multi ./build/dev-client ./src/main.js
```

iTurbine1:
I tried this template in NODE_ENV=development, testing and production with no issues:
[browserify](https://github.com/vuejs-templates/browserify)

Lacroute:
`NODE_ENV=development` fixed the problem for me too.

Turqueso:
Having same problem, @Lacroute where did you add `NODE_ENV=development`

Lacroute:
@Turqueso From command line `set NODE_ENV=development` and if you're on windows you should take a look at your environment variables ;)

kerf007:
Problem fixed for me too: `NODE_ENV=development`.
I added it in 'dev' script (for windows): `"dev": "cross-env NODE_ENV=development node build/dev-server.js"`

Volcanic-Penguin:
If an unmodified webpack template is giving errors I'd rather just use webpack-simple

moshie:
[...]

PR #1261, opened 2018-01-23
Add ExtractTextPlugin to webpack dev config

mrewers:
Upon running `npm run dev` the development build fails and returns the following error in reference to both src/App.vue and src/components/HelloWorld.vue:

```
Module build failed: Error: "extract-text-webpack-plugin" loader is used without the corresponding plugin, refer to https://github.com/webpack/extract-text-webpack-plugin for the usage example
```

Copying the the extract text plugin from webpack.prod.conf.js  to webpack.dev.conf.js rectifies the problem and allows the dev build to complete without errors.

### DIFF
--- a/template/build/webpack.dev.conf.js
+++ b/template/build/webpack.dev.conf.js
@@ -8,6 +8,7 @@ const baseWebpackConfig = require('./webpack.base.conf')
 const CopyWebpackPlugin = require('copy-webpack-plugin')
 const HtmlWebpackPlugin = require('html-webpack-plugin')
 const FriendlyErrorsPlugin = require('friendly-errors-webpack-plugin')
+const ExtractTextPlugin = require('extract-text-webpack-plugin')
 const portfinder = require('portfinder')
 
 const HOST = process.env.HOST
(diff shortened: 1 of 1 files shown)


LinusBorg:
The development setup uses style-loader, it doesn't extract css on purpose because we want hot reloading to work for styles. 


The error you are experiencing likely has to do with a misconfigured `NODE_ENV` environment variable. 

Open an issue and describe your problem.

mrewers:
Understood. I reverted the changes locally and set my environment variables to development. It took a little trial and error to get the right command (I'm on a Windows machine) but eventually `$env:NODE_ENV="development"` worked. Thank you for your help.

LinusBorg:
You should not have this env variable set permanently on your machine. 

The project commands will choose the right default value for you.

mrewers:
Okay, makes sense. I will remove the global env variable and figure out how to set the process env variable instead. Essentially, my problem is described in issue #535 but in reverse. When I have some time I'll see if I can find some answers in that thread. Or maybe I need to add something like cross-env. If I can't figure it out I'll open a new issue describing my problem more fully. Thanks again and sorry for the misguided pull request.

PR #1323, opened 2018-03-04
Simplify code in utils.js

dima74:


LinusBorg:
That syntax isn't supported in node 6, which is the minimum node version this template intends or to support. 

I can't merge this for that reason.

PR #1324, opened 2018-03-07
remove logo from build folder

c0defre4k:
I don't think this is supposed to be there?


LinusBorg:
It is supposed to be there. It's used in the error notifications if they are activated.

c0defre4k:
😯ah, i must have overlooked that. Sorry!

Issue #1367, opened 2018-04-25
Update to vue-loader v15?


kuoruan:
Evan You released vue-loader v15, the config is different from v 14, when will this template update?

https://vue-loader.vuejs.org/migrating.html

ChadTaljaardt:
Please follow the thread here: 

https://github.com/vuejs-templates/webpack/issues/1280

kuoruan:
OK, Thx

Issue #1280, opened 2018-01-30
Webpack 4


adam-lynch:
Hey, [Webpack 4 will be released soon](https://medium.com/webpack/webpack-4-beta-try-it-today-6b1d27d7d7e2). I'm creating this issue to track when this template is updated so I can update my projects. And I'm particularly interesting in how #1022 will be affected (or not).

michaelgallaghertw:
for quick reference, the changelogs:

* https://github.com/webpack/webpack/releases/tag/v3.0.0
* https://github.com/webpack/webpack/releases/tag/v4.0.0-beta.0

In theory there are no backwards compatibility breaks for v3. Points to watch out for from v4:

* CommonsChunkPlugin was removed -> optimization.splitChunks
* import() always returns a namespace object. CommonJS modules are wrapped into the default export
    * I'd expect some fallout from this, possibly within the `inject-loader`

KaelWD:
It's out now: https://github.com/webpack/webpack/releases/tag/v4.0.0

CharlesKumar:
Compelling reason for upgrading

1. Better Defaults — #0CJS (Zero configuration - no need for a webpack config file)
2. webpack takes care of dead code elimination (previously done by uglifyjs)
3. huge performance/bundle size improvement over webpack 3.

Things to consider

1. webpack-cli is separated
2. webpack 4 has newer plugin system (current webpack 3 plugin system deprecated)
3. changes in common js importing 

> Syntax

    import() always returns a namespace object. CommonJS modules are wrapped into the default export
        This probably breaks your code, if you used to import CommonJs with import()


4. changes in module type support (auto, esm, json, webassembly)

> Defaults

    webpack now looks for the .wasm, .mjs, .js and .json extensions in this order


nodejs 4 support is dropped (expected)

 a fork of vuejs-templates/webpack for webpack 4 (until everyone realizes webpack 4 has huge benefits over webpack 3)

jzs11:
Not a smooth upgrade with Webpack 4 particular extract-text-webpack-plugin
https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/727

CharlesKumar:
As I mentioned, webpack 4 has new plugin system

The extract-text-plugin is currently in alpha version (there's no stable release for webpack 4 yet)

plugins are required to rewriiten with webpack 4 compatible API 

plugin users have to keep an eye on releases and issues for smooth update (wait for the right time)

[...]

PR #1314, opened 2018-02-25
Remove mapCoverage as it is deprecated

termosa:
mapCoverage has been deprecated since Jest@22.4.0
https://github.com/facebook/jest/blob/master/CHANGELOG.md#2240
https://github.com/facebook/jest/pull/5177

PR #1369, opened 2018-04-27
webpack 4, vue-loader 15, et al

noamkfir:
This PR includes the following:
* `webpack` updated to 4.6.0
* `vue-loader` updated to 15.0.3
* many other packages updated to their most recent versions, including: `karma`, `mocha`, `sinon-chai`, `babel`, various `eslint` and `webpack` plugins and loaders, and others
* `extract-text-webpack-plugin` replaced with `mini-css-extract-plugin`

When upgrading `vue-loader`, I followed the [migration guide](https://vue-loader.vuejs.org/migrating.html). The notable exception is that I did not add CSS modules because that would require a bit of a rewrite of how the webpack rules are generated in `utils`, and CSS modules aren't supported yet by the template anyway, as far as I can tell.

Vue is kind of new for me, so I'm not sure how to verify that everything works correctly. Things seem to be working well enough for me :). The options I use are: standalone, with vue-router, eslint, and unit tests, the Standard esline preset, karma and mocha, Nightwatch, and npm. I've run all the npm scripts and they all seem to be doing their jobs.

#1280 
#1367


ChadTaljaardt:
Thanks for taking the time to do this.

noamkfir:
I'm trying to get the build to pass. I noticed that the chromedriver being used for the e2e tests is 2.35, which is apparently installed by the image in /usr/bin/chromedriver. It should be using 2.38, which is specified in package.json and located in node_modules.

The Dockerfile used for the build tries to [install the latest chromedriver](https://github.com/CircleCI-Public/circleci-dockerfiles/blob/d235dc7c366a318afdec7883be77185a32219a72/node/images/6.12.3/browsers/Dockerfile#L64), which should be 2.38, but either the script isn't working properly, some cache (maybe CircleCI's?) is returning the older version, or the chromedriver is preinstalled and the script is failing to overwrite it.

I'm trying to configure a build that is identical to the current build, except that it does not try to install the global chromedriver, by commenting out just the chromedriver installation command. Turns out it's pretty complicated to get that set up and running. I'm working on it, but if anybody knows a faster way to get CircleCI to use a slightly different Dockerfile, I would be obliged :).

noamkfir:
# Finally got the build to pass!

This mainly involved replacing the prebuilt CircleCI image with a custom image generated using [a fork of CircleCI's dockerfile-wizard](https://github.com/noamkfir/dockerfile-wizard).

The custom image is currently necessary because the prebuilt images install [chromedriver globally](https://github.com/CircleCI-Public/circleci-dockerfiles/blob/676d62b3c6348c39937c9a1bc5a61da02cc00053/node/images/6.12.3/browsers/Dockerfile#L64). For some reason, they install 2.35 instead of 2.38 (not sure why...). The global executable takes precedence over node_modules, causing the build to fail. I think my solution is better: I got rid of the global chromedriver installation altogether. package.json is now the single source of truth for the chromedriver version.

I also had to add the `--headless` arg to the `chromeOptions` property in nightwatch.json, as well as make a few other tweaks. Took me a while to figure out why the container couldn't launch Chrome, and Chrome's and chromedriver's inane error messages weren't very helpful. It was a combination of the browser, webdriver, selenium, and nightwatch versions, each with their own quirks and breaking changes.

I created a logs folder that the webdrivers write to. CircleCI collects those logs and puts them on the Artifacts tab after a build. Really helps figure out what's going on.

I'm not sure if this was necessary, but I upgraded to the latest Node 6 image that CircleCI explicitly supports - 6.14.1. I'm guessing Node 8 (and maybe even 10) should work, but I'm not sure why it was downgraded to 6 and didn't want to screw things up now that it finally works...

template/build/webpack.prod.conf.js
```diff
@@ -27,30 +28,16 @@ const webpackConfig = merge(baseWebpackConfig, {
...
-          warnings: false
-        }
-      },
-      sourceMap: config.build.productionSourceMap,
-      parallel: true
-    }),
     // extract css into its own file
-    new ExtractTextPlugin({
+    new MiniCssExtractPlugin({
       filename: utils.assetsPath('css/[name].[contenthash].css'),
```
gplusdotgr: if you don't provide an explicit chunkFileName here you'll get errors when you split several chunks for lazy loading. (something like trying to emmit to the same file)

template/build/utils.js
```diff
@@ -31,7 +31,22 @@ exports.cssLoaders = function (options) {
 
   // generate loader string to be used with extract text plugin
   function generateLoaders (loader, loaderOptions) {
-    const loaders = options.usePostCSS ? [cssLoader, postcssLoader] : [cssLoader]
+    const loaders = []
+
+    // Extract CSS when that option is specified
+    // (which is the case during production build)
+    if (options.extract) {
```
gkatsanos: I don't see this passed/used somewhere?

gplusdotgr:
Have you checked that SourceMaps work? (because I think there are issues with mini-css-extract-plugin and sourceMaps)

template/build/vue-loader.conf.js
```diff
@@ -1,19 +1,9 @@
 'use strict'
```
afontcu, line 1: Hi! Does this file still makes sense? vue-loader config was weird enough to justify its own file, but since now most of its configuration is wiped, should we consider inlining it to `webpack.base.conf.js`?

template/build/webpack.prod.conf.js
noamkfir: Thank for looking over the code @gkatsanos!

Are you sure, though? The [readme](https://github.com/webpack-contrib/mini-css-extract-plugin) says both `filename` and `chunkFilename` are optional. And the [webpack docs](https://webpack.js.org/configuration/output/#output-chunkfilename) say:
> By default `[id].js` is used or a value inferred from `output.filename` (`[name]` is replaced with `[id]` or `[id].` is prepended)

I'm assuming it's inferred from `filename` in this case, but haven't verified.

If it does break that way, how would I repro this effectively? I'd probably need more than just a long CSS file to test, right?

In any case, I was trying to mirror the configuration used for the `extract-text-webpack-plugin`, and this combination seemed to produce correct results (eyeballed). It's possible we need to reconsider the default settings.

gplusdotgr: Try having 2 or 3 lazy loaded components using vue-router when each one's using a separate CSS file.

template/build/vue-loader.conf.js
noamkfir: @gkatsanos, they seem to work well on my machine... The CSS file in `dist` is minified and when I click a selector, Chrome opens the `.vue` file and highlights the proper line.

Can you point me to something specific so I can verify?

noamkfir: @afontcu, I think it should stay, but I'm not really sure.

Vue is very new for me, so I don't know how every setting affects the build or the app. I only made changes I felt certain were equivalent to the previous behavior or completely redundant, mainly by following Vue's [migration guide](https://vue-loader.vuejs.org/migrating.html), reading docs, and trial and error.

For example, I got rid of the `loaders` property because the new vue-loader explicitly deprecates it in favor of the the webpack loaders configuration. And I renamed the `transformToRequire` property to `transformAssetUrls`, based on the guide.

My hunch is that it should probably stay for the sake of clarity, even though the usage has changed a bit.

Having said that, I haven't checked if the `cacheBusting` property is used anywhere, and I'm not sure if we should add other properties that are relevant for the new vue-loader.

ChadTaljaardt:
What still needs to be done with this before it can be merged?

template/build/webpack.prod.conf.js
noamkfir: @gkatsanos Finally found some time to take another look at this. I have to say, you sent me down the rabbit hole with this one...

First of all, I was able to determine relatively quickly that if `filename` is specified and `chunkFilename` *is not*, webpack uses `filename` by default.

Second, I'm pretty sure we don't need `[contenthash]`, which was used by the `extract-text-webpack-plugin`. The `[chunkhash]` is probably more suitable for our purposes anyway.

However, that's not the end of the story. Turns out that regardless of the naming strategy used here, the current configuration may generate a new `chunkhash` for the `vendor` chunk for any change in the code, even if it is not in `node_modules`. **This is probably a blocker.**

I did a bit of research and played around with different variations of `filename` and `chunkFilename` for both `output` and the `MiniCssExtractPlugin` using different combinations of placeholders (`[name]`, `[id]`, `[contenthash]` and `[chunkhash]`). I also tested with and without explicitly named chunks in dynamic imports for routing (e.g., `const SampleComponent = () => import(/* webpackChunkName: "sample" */ '@/components/SampleComponent')`, which imports a `SampleComponent.vue` file). And I fiddled quite a bit with the `optimization` settings.

**The bottom line is that I've solved it.** I have to clean it up a bit, so I hope to push it in the next day or two.

noamkfir:
@ChadTaljaardt, following up on @gkatsanos's review, I noticed that the `vendor` chunk does not preserve its `chunkhash`, so the PR probably shouldn't be merged just yet.

I've fixed it and will try to push the update in the next day or two.

noamkfir:
This update:
* simplifies the config by resorting to some default and recommended webpack 4 settings
* unifies the file and chunk naming strategies for js and css files for easier debugging and improved long-term caching
* stabilizes the chunk hashes, mainly to prevent generating a new chunk hash for the vendor chunk when only app code is modified, but applied generally

If all is well, this should hopefully remove any remaining obstacles to merging.

ChadTaljaardt:
Hello @noamkfir 

Im not sure if this is caused by your implementation but i'm getting this when i run 
```
npm run test
```

```
● Deprecation Warning:

  Option "mapCoverage" has been removed, as it's no longer necessary.

  Please update your configuration.

  Configuration Documentation:
  https://facebook.github.io/jest/docs/configuration.html

```

ChadTaljaardt:
Sorry i just realised it was in this PR : https://github.com/vuejs-templates/webpack/pull/1314

gplusdotgr (approved):

template/test/e2e/nightwatch.conf.js
```diff
@@ -1,45 +1,93 @@
...
 var config = require('../../config')
 
+const paths = {
+  chromedriver: require('chromedriver').path,
+  geckodriver: require('geckodriver').path,
+  seleniumServer: require('selenium-server').path,
+  launchUrl: 'http://localhost:' + (process.env.PORT || config.dev.port),
+  logPath: 'logs',
+}
+console.log('paths:', paths)
```
dak: Is this debug code?

noamkfir: The `paths` object is not debug code. I intentionally collected all the paths for readability and to simplify the configuration.

The `console.log` is debug code and is extremely useful when trying to figure out why the webdrivers aren't loading.

I just pushed a small update (ad1999d) that:
* adds a debugTests constant to `nightwatch.conf.js`
* logs the `paths` object when it's `true`
* sets the nightwatch config object's `silent` property to `!debugTests`

I also cleaned up the file a bit.

ChadTaljaardt (approved):

Dylan-Chapman:
Any plans for vue-cli 3 in this?

MehmetKursat:
I can't wait for this to be merged, also I wonder if I use the old version and then the new version of this comes, will I have to easily upgrade with npm install or something or is it better to wait for this to be released before i start on a new vue project?

jjyyxx:
@noamkfir 
I'm sorry to report some issues of the template. This may not connect directly with your template, instead, connect with those plugins, but I'm seeking for your help (and anyone's, of course).

The current template has severe **memory leak** in dev mode. The memory keeps going up every time the dev server recompiled the files. Since my project is relatively big, the memory quickly exceeds the heap after around 10 compilation and exits with code 134.

I did memory dump with chrome-devtool's node debug feature. Four snapshots are taken after four minimum recompilation (remove a statement/add it back).
image
I have saved those snapshots. If you think it necessary and would like to analyze it, you can download it from google drive and load it from chrome-devtool.
[snapshot 1](https://drive.google.com/open?id=1QPbXHlhX-s1cgZNF05YH6yXbRA-AkIxt)
[snapshot 2](https://drive.google.com/open?id=1pWHhAkmD1WyUBl6X5ArCG1k5W-wpTOPc)
[snapshot 3](https://drive.google.com/open?id=1ZmZAST8ew7jKt4_6y49T5DVt8GsyJgVX)
[snapshot 4](https://drive.google.com/open?id=1sbeYkN8tVAb7zMo9XLa770SLUsAqGcII)
If you need any further information, I'm glad to provide.

BTW, I updated the devDependencies to newest:
```javascript
{
  "devDependencies": {
    "autoprefixer": "^8.5.0",
    "babel-core": "^6.26.3",
    "babel-eslint": "^8.2.3",
    "babel-helper-vue-jsx-merge-props": "^2.0.3",
    "babel-loader": "^7.1.4",
    "babel-plugin-syntax-jsx": "^6.18.0",
    "babel-plugin-transform-runtime": "^6.23.0",
    "babel-plugin-transform-vue-jsx": "^3.7.0",
    "babel-preset-env": "^1.7.0",
    "babel-preset-stage-2": "^6.24.1",
    "chalk": "^2.4.1",
    "copy-webpack-plugin": "^4.5.1",
    "css-loader": "^0.28.11",
    "eslint": "^4.19.1",
    "eslint-config-standard": "^11.0.0",
    "eslint-friendly-formatter": "^4.0.1",
    "eslint-loader": "^2.0.0",
    "eslint-plugin-import": "^2.12.0",
    "eslint-plugin-node": "^6.0.1",
    "eslint-plugin-promise": "^3.7.0",
    "eslint-plugin-standard": "^3.1.0",
    "eslint-plugin-vue": "^4.5.0",
    "file-loader": "^1.1.11",
    "friendly-errors-webpack-plugin": "^1.7.0",
    "html-webpack-plugin": "^3.2.0",
    "mini-css-extract-plugin": "0.4.0",
    "node-notifier": "^5.2.1",
    "optimize-css-assets-webpack-plugin": "^4.0.1",
    "ora": "^2.1.0",
    "portfinder": "^1.0.13",
    "postcss-import": "^11.1.0",
    "postcss-loader": "^2.1.5",
    "postcss-url": "^7.3.2",
    "rimraf": "^2.6.2",
    "semver": "^5.5.0",
    "shelljs": "^0.8.2",
    "uglifyjs-webpack-plugin": "^1.2.5",
    "url-loader": "^1.0.1",
    "vue-loader": "^15.2.0",
    "vue-style-loader": "^4.1.0",
    "vue-template-compiler": "^2.5.16",
    "webpack": "^4.8.3",
    "webpack-bundle-analyzer": "^2.13.1",
    "webpack-cli": "2.1.4",
    "webpack-dev-server": "^3.1.4",
    "webpack-merge": "^4.1.2"
  }
}
```

ChadTaljaardt:
@jjyyxx I am getting this too, but I don't think it has to do with this template.

Here is another issue I am following in regards to this. https://github.com/webpack/webpack/issues/6929

jjyyxx:
Thanks for your reference! Actually I have seen that issue.

What surprised me was that I configured a similar `vue-webpack4` project recently and didn't encounter memory leaks (at least, wasn't aware of memory leaks). I switched to this template mainly because my own template has a strange bug when configuring dev-server and copy-webpack-plugin (`contentbase` to be more specific). This template solved my problem but led to a new one...

The differences between my configuration and this one are quite small. Maybe a specific option triggered the problem.

ChadTaljaardt:
@jjyyxx If your configuration doesn't leak, and this configuration does, would you be able to share your current config so that we might be able to investigate where it could be? It must be in where there is a difference between the config's which makes it easier to narrow down where the issue is.

jjyyxx:
@ChadTaljaardt 

I investigated the previous configuration, and found that replacing the lines in `webpack.dev.conf.js`
```
rules: utils.styleLoaders({ sourceMap: config.dev.cssSourceMap, usePostCSS: true })
```
to
```
rules: [
  {
    test: /\.css$/,
    use: [
      'vue-style-loader',
      'css-loader'
    ]
  }
]
```
significantly eased the leak. I know it isn't a fix as it drops support for css dialects, but it turned out that something went wrong in the `styleLoaders` function or the relating loaders.

The mem dump:
mem dump

Description: Between Row 1 ~ Row 5, one minimum recompilation each time. Between Row 5 and Row 6, invoke a GC operation. So, it seems that other minor leaks exist, but it isn't so severe.

I would be appreciated if anyone can trace into the `styleLoaders` and locate the problem.

jjyyxx:
In complement, the return value of 
```javascript
utils.styleLoaders({ sourceMap: config.dev.cssSourceMap, usePostCSS: true })
```
is
```javascript
[
    {
        test: /\.css$/,
        use:
            ['vue-style-loader',
                {loader: 'css-loader', options: {sourceMap: true}},
                {loader: 'postcss-loader', options: {sourceMap: true}}]
    },
    {
        test: /\.postcss$/,
        use:
            ['vue-style-loader',
                {loader: 'css-loader', options: {sourceMap: true}},
                {loader: 'postcss-loader', options: {sourceMap: true}}]
    },
    {
        test: /\.less$/,
        use:
            ['vue-style-loader',
                {loader: 'css-loader', options: {sourceMap: true}},
                {loader: 'postcss-loader', options: {sourceMap: true}},
                {loader: 'less-loader', options: {sourceMap: true}}]
    },
    {
        test: /\.sass$/,
        use:
            ['vue-style-loader',
                {loader: 'css-loader', options: {sourceMap: true}},
                {loader: 'postcss-loader', options: {sourceMap: true}},
                {
                    loader: 'sass-loader',
                    options: {indentedSyntax: true, sourceMap: true}
                }]
    },
    {
        test: /\.scss$/,
        use:
            ['vue-style-loader',
                {loader: 'css-loader', options: {sourceMap: true}},
                {loader: 'postcss-loader', options: {sourceMap: true}},
                {loader: 'sass-loader', options: {sourceMap: true}}]
    },
    {
        test: /\.stylus$/,
        use:
            ['vue-style-loader',
                {loader: 'css-loader', options: {sourceMap: true}},
                {loader: 'postcss-loader', options: {sourceMap: true}},
                {loader: 'stylus-loader', options: {sourceMap: true}}]
    },
    {
        test: /\.styl$/,
        use:
            ['vue-style-loader',
                {loader: 'css-loader', options: {sourceMap: true}},
                {loader: 'postcss-loader', options: {sourceMap: true}},
                {loader: 'stylus-loader', options: {sourceMap: true}}]
    }
]
```
I did not use any css dialect so maybe it's something related to `postcss-loader`?

noamkfir:
@jjyyxx I was able to reproduce the leak very easily, so I dug into it a bit and I'm pretty sure this isn't related to my changes.

Basically, I discovered that the leak also occurs in the released version of the template. As far as I can tell, it's the exact same leak, almost certainly the one that @ChadTaljaardt referenced: webpack/webpack#6929.

I don't think it's the `postcss-loader` plugin because removing it does not seem to prevent the leak, but I can't rule it out either yet. It's entirely possible that a number of plugins are incorrectly retaining references to the `Compilation` and/or `CachedSource` objects between builds.

I tried to replace `module.rules` as you suggested and this also didn't seem to prevent the leak. I'm not sure why I'm getting different results than you are but I tried a number of different approaches and still wasn't able to reproduce the improvement.

One thing I found interesting was that there's a difference in the memory profile between webpack-dev-server and webpack. It's not completely deterministic, but webpack seems to leak less and more predictably. This is true for both dev and prod modes.

jjyyxx:
With so many plugins, tracing the root of the leak is really tricky.
Thanks for your efforts anyway @noamkfir. Hope this problem can be solved soon.

gplusdotgr:
Since this is not related to the pull request what's stopping us from merging it?

jamesjieye:
For anyone who cannot wait and want to try webpack4, use `vue init noamkfir/webpack <app name>` to bootstrap your app. **Not for production**

lucileFievet:
@jamesjieye ,

I tried your command :

>  vue init noamkfir/webpack **copy**_of_my_old_application

on my old application **copy** and I get this error message : 

>  **Module build failed:** 
> @import './style.scss"' in myvuefile.vue

This problem can be **solved** by setting a **sccs  filter after js one** in webpack.base.conf.js

```
include: [resolve('src'), resolve('test'), resolve('node_modules/webpack-dev-server/client')]
},
{
+        test: /\.scss$/,
+        include: [resolve('src')],
+        use: [
+          'vue-style-loader',
+          'css-loader',
+          'scss-loader',
+          'sass-loader'
+        ]
+      },
+      {
test: /\.(png|jpe?g|gif|svg)(\?.*)?$/,
```
_I also tried to set the include: resolve('src') in utils.js  but it didn't worked ..._

```
 exports.styleLoaders = function (options) {
     const loader = loaders[extension]
     output.push({
       test: new RegExp('\\.' + extension + '$'),
+      include: path.join(__dirname, '..', 'src'),
       use: loader
     })
   }

```
Now,  it seems I have a problem with components called by vue-rooting or with props ... ?

> These relative modules were not found:
> 
> * ./Checkbox.vue?vue&type=style&index=0&id=1f980e0b&scoped=true&lang=scss in ./src/components/formElements/checkbox/Checkbox.vue

the Checkbox.vue :

```
<style scoped  lang="scss">
@import '../../../assets/vars/colors.scss';
@import '../../../assets/mixins/checkbox.scss';
</style>
<template>
    <label
        class="b-checkbox checkbox"
        :class="[size, { 'is-disabled': disabled }]"
        ref="label"
        :disabled="disabled"
        :tabindex="disabled ? false : 0"
        @keydown.prevent.enter.space="$refs.label.click()">
        <input
            v-model="value"
            :id="id"
            type="checkbox"
            :disabled="disabled"
            :name="name"
            :value="nativeValue"
            :true-value="trueValue"
            :false-value="falseValue">
        <span :class="checkClass"/>
        <span class="control-label"><slot/></span>
    </label>
</template>

<script>
export default {
  name: 'Checkbox',
  props: {
    id: [String, Number],
    nativeValue: [String, Number, Boolean, Function, Object, Array, Symbol],
    disabled: Boolean,
    name: String,
    size: String,
    trueValue: {
      type: [String, Number, Boolean, Function, Object, Array, Symbol],
      default: true
    },
    falseValue: {
      type: [String, Number, Boolean, Function, Object, Array, Symbol],
      default: false
    },
    type: {
      type: String,
      default: 'is-primary'
    }
  },
  data () {
    return {
      newValue: this.nativeValue,
      value: this.nativeValue,
      checkClass: this.initClass()
    }
  },
  methods: {
    initClass: function () {
      // console.log('initClass native val: ' + this.nativeValue + ', true val : ' + this.trueValue)
      if (this.nativeValue === this.trueValue) {
        return this.type + ' check checked'
      } else {
        return this.type + ' check unchecked'
      }
    },
    check: function () {
      this.checkClass = this.type + ' check checked'
    },
    uncheck: function () {
      this.checkClass = this.type + ' check unchecked'
    }
  },
  watch: {
    /**
     * When v-model change, set internal value.
     */
    value (value) {
      this.newValue = value
    },
    /**
     * Emit input event to update the user v-model.
     */
    newValue (value) {
      // console.log('newValue val: ' + value + ', true val : ' + this.trueValue)
      this.$emit('input', {id: this.id, name: this.name, checked: value})
      if (value === this.trueValue) {
        this.check()
      } else {
        this.uncheck()
      }
      // this.$parent.changeValue(this.id, value)
    }
  }
}
</script>
```

jamesjieye:
Hi @lucileFievet, 

I'm able to create a brand new application with `vue init noamkfir/webpack <app name>`. I didn't try it on existing application.

lucileFievet:
Hy @jamesjieye ,

Yes and it does work with default application. My code works with webpack 3.9... 
Finaly I don't think it is the router or props ... 

I think it is the way I import external sccs in scoped  style. It doesn't like the @import in style tag. 
(It is worse with vue-cli  :/) 
I 'll try a work arround.
Thanks anyway

miaulightouch:
I update this pr to vue-cli 3, babel 7: https://github.com/miaulightouch/webpack

and I post the pr here: https://github.com/noamkfir/webpack/pull/1

try it now: `vue init miaulightouch/webpack#develop <project name>`

noamkfir:
Thanks @miaulightouch! I've merged your PR into mine.

template/package.json
```diff
@@ -49,73 +49,76 @@
...
-    "vue-style-loader": "^3.0.1",
+    "url-loader": "^1.0.1",
+    "vue-loader": "^15.0.3",
+    "vue-style-loader": "^4.1.0",
     "vue-template-compiler": "^2.5.2",
     "portfinder": "^1.0.13",
-    "webpack": "^3.6.0",
-    "webpack-dev-server": "^2.9.1",
+    "webpack": "^4.6.0",
+    "webpack-cli": "^3.0.3",
```
miaulightouch, line 121: is `webpack-cli` required for vue template?

xudafeng:
https://github.com/ant-ife/vue-biz-app-template/pull/5

aksenovdev:
Hello, can someone check tests of project? I've some error when try run `npm test` in clean @miaulightouch template
[debug.log](https://github.com/vuejs-templates/webpack/files/2112306/debug.log)
[console.log](https://github.com/vuejs-templates/webpack/files/2112305/console.log)


Error:
image

ChadTaljaardt:
@aksenovdev I tested this with the pr version and it seems to be working fine. 
```
vue init noamkfir/webpack test
```

then ran 
```
cd test
```
 then ran
```
npm test
```

miaulightouch:
@aksenovdev you can check my CI report [here](https://circleci.com/gh/miaulightouch/webpack/tree/develop)

huangsam:
Encountered the same issue as @aksenovdev with two test runners:

Jest:

screen shot 2018-06-18 at 10 44 50 pm

Karma + Mocha:

screen shot 2018-06-18 at 10 54 00 pm

Errors were the same regardless of running `npm run test` or `yarn run test`.

**Note**: I ran `vue init noamkfir/webpack test` to initialize the project, once for each of the test runners I described above. This was run on a Macbook Pro 2017 with Node 10.4.1 installed via Homebrew.

miaulightouch:
`@babel/runtime` is missing for now. please  install `@babel/runtime` to resolve this issue, I'm inspecting the problem, maybe this is from upstream changes.

aksenovdev:
@miaulightouch Yea, that helped, thank you)

miaulightouch:
`babel-core@6` include `babel-runtime` as dependency, but `@babel/core`(babel 7) not.
also, my CI keep `@babel-runtime` in workspace from unknown source (maybe previous builds?) and make all test passed, then only new created project would encounter this issue.

now, this template would include `@babel/runtime` if [this pr](https://github.com/noamkfir/webpack/pull/2) get merged.
@noamkfir

jacobg:
Does this work include the pwa support in vue cli 3.0?

noamkfir:
Sorry about the delay. Been a bit busy. I've merged @miaulightouch's PR.

noamkfir:
@jacobg and @Dylan-Chapman This PR does **not** include the PWA support added in Vue CLI 3. In fact, I'm not really sure how the new CLI will affect this project. As I see it, the new CLI is more of a paradigm shift than a rewrite. Someone with more experience than I have will have to weigh in on whether this project is even relevant for CLI 3 projects.

This PR essentially updates `webpack` and `vue-loader`. It's a pretty big PR because so many packages had to be updated, underlying behavioral changes accommodated, and configurations reconfigured, but a lot of effort was made to avoid changing surface behavior.

template/package.json
noamkfir: I'm not entirely sure if `webpack-cli` is *required*, but all the migration guides I looked at said to install it. The CLI code was split out from `webpack` in v4 and I wasn't really sure how to check if the generator or the templates use any of its functionality anywhere. I had planned to look into it further, but never got around to it.

LinusBorg:
> In fact, I'm not really sure how the new CLI will affect this project. As I see it, the new CLI is more of a paradigm shift than a rewrite. Someone with more experience than I have will have to weigh in on whether this project is even relevant for CLI 3 projects.

That's right. vue-cli 2 was basically just pulling templates like this webpack template from github.

vue-cli 3 doesn't rely on project templates anymore. Instead, it extracted all the config and dependencies for webpack, eslint and so forth into npm packagesthat cn ge integrated into any vue-cli 3 project as plugins.

Plugins can of course also be written by the community. They can alter the project's configuration, ad dependencies, add files from templates, add npm script commands and much more.

That means that there's not really a need for templates such as this webpack template or the pwa template. Instead, everyone can mix and match the feature they need by installing the corresponding plugins.

So what does that mean for the future of this template? Personally, I don't see  reason to start a new project with this template anymore. vue-cli comes with all of its features and then some, and it has the advantage of being easily upgradable, which this template isn't.

But nonetheless, thanks to @noamkfir for the work on this PR. I will hopefully merge it soon, but that will be the last bigger change that I will make to this template. As a consequence of my explanations about vue-cli 3, I don't see it as worthwhile to put much time into it anymore.

But if someone want to take over maintanance, we could certainly talk about this. And there's always the possibility to fork this project :)

Garito:
Hi!
I should be agree with @LinusBorg about the point of this template
But!
That doesn't mean that this ends here
As I see it you have a good amount of knowledge stacked since this project starts
Why don't use it to start small plugins that solves the same features but with the new cli?
I bet you that you will come up with ideas as soon as you start to check what has been made so far
I would like to thank you, too, for your work till now (I've started and not gone back till cli3 with the pwa which uses this one as base)

LinusBorg:
@Garito every single feature of this template is already included in vue-cli 3.

And if you do a quick search on npmjs.com, you will find that there are already dozens of plugins available for it, including things like vuetify, i18n, Apollo graphql etc. pp.

Garito:
I know but there's always something to do
I'm trying not to be negative only while expressing my agreement with the idea of the obsolescence of this template
Isn't easy to accept sometimes (we have jumped from project to project several times)
But always that I've changed my path because of this I carry new skills which help me start over
I just can thank you

mikeerickson:
As much as I applaud and like projects such as the Vue CLI, there is always room for options. I my case, we have taken the approach of iterating much of our UI forms from PHP to more API based and all new UI work is JS (we have used Angular, React and Vue)

template/build/webpack.prod.conf.js
```diff
@@ -120,7 +75,32 @@ const webpackConfig = merge(baseWebpackConfig, {
...
+  ],
+  optimization: {
+    concatenateModules: true,
+    splitChunks: {
+      chunks: 'all',
+      cacheGroups: {
+        vendor: {
+          name: 'vendor',
+          test: /[\\/]node_modules[\\/]/,
+          enforce: true,
```
zevdg, line 86: I can't find any documentation on this enforce option? 
https://webpack.js.org/plugins/split-chunks-plugin/#splitchunks-cachegroups

noamkfir: @zevdg Quoting @thatisuday (https://github.com/webpack/webpack/issues/6647#issuecomment-395078260):

> `enforce` will force **modules** to split whether or not condition mentioned in `splitChunks` suitable to modules to split.

If I recall correctly, I added it to ensure consistent hashing, but I really don't remember the details and, in hindsight, that probably doesn't make much sense :(. I do remember that it took me way too long to solve whatever the problem was because, as you said, there was no documentation for `enforce` so I didn't know the option existed until I came across it by accident on some unrelated article.

The [SplitChunksPlugin code](https://github.com/webpack/webpack/blob/master/lib/optimize/SplitChunksPlugin.js#L264) basically uses it to arbitrarily set other parameters to default values that guarantee that the chunk will be created.

HenryWong-fe:
@LinusBorg  I know that vue-cli has a lot of configuration items, but how can vue-cli ensure that my project public files can exist every time I create them? I need to use some configuration of the meta files in the template to decide whether to use some plugins and file

template/build/webpack.dev.conf.js
```diff
@@ -65,7 +64,11 @@ const devWebpackConfig = merge(baseWebpackConfig, {
         ignore: ['.*']
       }
     ])
-  ]
+  ],
+  optimization: {
```
PanJiaChen, line 68: This is not necessary. In the `mode: 'development'` is the default configuration.

template/build/webpack.prod.conf.js
```diff
@@ -75,43 +61,12 @@ const webpackConfig = merge(baseWebpackConfig, {
         // more options:
         // https://github.com/kangax/html-minifier#options-quick-reference
       },
-      // necessary to consistently work with multiple chunks via CommonsChunkPlugin
+      // necessary to consistently work with multiple chunks
       chunksSortMode: 'dependency'
     }),
     // keep module.id stable when vendor modules does not change
+    new webpack.NamedChunksPlugin(),
```
PanJiaChen, line 68: https://webpack.js.org/guides/caching/#module-identifiers

`HashedModuleIdsPlugin` is enough

template/build/webpack.prod.conf.js
```diff
@@ -120,7 +75,32 @@ const webpackConfig = merge(baseWebpackConfig, {
         ignore: ['.*']
       }
     ])
-  ]
+  ],
+  optimization: {
+    concatenateModules: true,
```
PanJiaChen: In  `mode: 'production'`, it default enable.

template/build/webpack.prod.conf.js
noamkfir: @PanJiaChen Thanks for the thorough review!

I did a lot of experimentation before I added the `NamedChunksPlugin` because the `HashedModuleIdsPlugin` did not guarantee consistent hashing, especially of the `vendor` module.

I found that almost every change I made to any `import` statement in my code could cause the `vendor` module hash to change. The `NamedChunksPlugin` uses the path instead of the arbitrary module id to calculate the hash, and this proved to be far more consistent.

It's definitely possible I missed something, or that the behavior improved since webpack 4.6 (it's now 4.16), but I hesitate to reopen this right now as changing this or upgrading to 4.16 might have further consequences. In any case, after this PR is approved, doing that kind of change is a much smaller and far more focused change.

template/build/webpack.prod.conf.js
noamkfir: 👍

template/build/webpack.dev.conf.js
noamkfir: You're right about `namedModules`. I'm removing it.

However, the [docs](https://webpack.js.org/configuration/optimization/#optimization-noemitonerrors) for `noEmitOnErrors` seem to indicate that it does not have a default value, so I'm leaving it for now.

template/build/webpack.prod.conf.js
```diff
@@ -27,30 +28,15 @@ const webpackConfig = merge(baseWebpackConfig, {
...
     // extract css into its own file
-    new ExtractTextPlugin({
-      filename: utils.assetsPath('css/[name].[contenthash].css'),
-      // Setting the following option to `false` will not extract CSS from codesplit chunks.
-      // Their CSS will instead be inserted dynamically with style-loader when the codesplit chunk has been loaded by webpack.
-      // It's currently set to `true` because we are seeing that sourcemaps are included in the codesplit bundle as well when it's `false`, 
-      // increasing file size: https://github.com/vuejs-templates/webpack/issues/1110
-      allChunks: true,
+    new MiniCssExtractPlugin({
+      filename: utils.assetsPath('css/[name].[chunkhash].css'),
```
PanJiaChen, line 39: Better with `contenthash`.

https://github.com/webpack-contrib/mini-css-extract-plugin/issues/135

yuicer:
well....will it be merged soon? or it still has some problems to do?
thanks to all of you ；）

ruchernchong (approved):

template/build/utils.js
```diff
@@ -31,7 +31,22 @@ exports.cssLoaders = function (options) {
 
   // generate loader string to be used with extract text plugin
   function generateLoaders (loader, loaderOptions) {
-    const loaders = options.usePostCSS ? [cssLoader, postcssLoader] : [cssLoader]
+    const loaders = []
```
ruchern: Shouldn't this be `let` rather than `const`?

template/package.json
ruchernchong: @miaulightouch `webpack-cli` is required to be installed otherwise you get an error message squawking.

template/build/webpack.prod.conf.js
```diff
@@ -120,7 +75,31 @@ const webpackConfig = merge(baseWebpackConfig, {
...
+        vendor: {
+          name: 'vendor',
+          test: /[\\/]node_modules[\\/]/,
+          enforce: true,
+        },
+      },
+    },
+    runtimeChunk: 'single',
+    minimizer: [
+      new UglifyJsPlugin({
```
ruchernchong, line 92: Is this needed since you are already using `mode: production`? 

According to the [docs](https://webpack.js.org/concepts/mode/#mode-production), they have removed `UglifyJsPlugin`.

ruchernchong:
Hi @LinusBorg , what is the status of this? It does looks good to merge now.

mjdietzx:
Any update here, would really like to updgrade to webpack4! Thanks for this PR!

PanJiaChen:
I think it is easier and safer to migrate directly to vue-cli@3 than to upgrade from this template to webpack4 now.

ruchernchong:
Old projects might still need this

mjdietzx:
@PanJiaChen i've been migrating my project to vue-cli@3 and it's amazing! so i actually don't need this anymore

miaulightouch:
some of deps still using beta package, I would take time for updating these packages.

and I would say in some case, like desktop app development, this template is still a good start.

template/build/utils.js
```diff
@@ -1,7 +1,7 @@
 'use strict'
 const path = require('path')
 const config = require('../config')
-const ExtractTextPlugin = require('extract-text-webpack-plugin')
+const MiniCssExtractPlugin = require('mini-css-extract-plugin')
 const packageConfig = require('../package.json')
```
mikestreety: Thanks so much for this PR, saved me a lot of pain. Very small thing - `packageConfig` doesn't get used in this file, so this throws an error in eslint 👍

LeFnord:
is where any progress on it, like to have mörged it, thanks

ruchernchong:
This PR is great, but Webpack 5 is imminent and I am not sure if this will come effective by then. 

Also, `vue-cli` is now v3 and this should be deprecated by now.

PR #1379, opened 2018-05-14
Delete redundant image file

caikan:


LinusBorg (commented):
This is not redundant. It's used by the notification feature.

PR #1394, opened 2018-05-31
 `vue init` command updated to `vue-init`

donnyquixotic:
`vue init` command when initializing project is incorrect and fails. The correct command is `vue-init` (or `sudo vue-init` if  root perm required).

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ This is a project template for [vue-cli](https://github.com/vuejs/vue-cli). **It
 
 ``` bash
 $ npm install -g vue-cli
-$ vue init webpack my-project
+$ vue-init webpack my-project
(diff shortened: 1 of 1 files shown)


LinusBorg:
You likely installed `@vue/cli`, which is the 3.0 version and a complete rewrite. 

It doesn't use templates anymore, and doesnt have an `init` command anymore, but a `create` command. 

The old, 2.* `init` command was extracted into a plugin package, called `@vue/cli-init`

Install it globally and `vue init webpack` will work again. But I recommend to just use the new `create` instead - the new CLI is in RC state already and much better than this template in many ways 

For further information, please read the docs of the new cli at cli.vuejs.org

PR #1398, opened 2018-06-06
delete unused css

manu020766:
delete unused css when run npm run build


LinusBorg:
This contains stuff we to vuetify, maybe this PR was a mistake?

aslam-c:
Hello,
I am using this template for an e-commerce like app.
and got trouble when try to **minify unused css **, (_putting all bootstrap.min.css contributes a large chunk to the final bundle_)
and the tutorials i got ,shows that i have to replace _extract-text-plugin_ with _mini-css-extract_ (_needs webpack 4.0^_),and when i did this, i got problems with this (webpack)compatibilty !

Is there any  other ways  ?

### i want to remove all unused css in final bundle inorder to reduce the size

PR #1438, opened 2018-08-24
加入vuex

wujun8008:


ruchernchong:
Translation: Add `Vuex` as a default package in the template.

